PR #4, opened 2019-10-10
Test-suite skeleton

rsvoboda:
Test-suite skeleton
 - usage described in README
 - downloads WF distribution if `jboss.home` is not defined
 - ARQ can connect to running server to speed up tests development turnaround
 - uses RestAssured
 - example module for MP Health

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,23 +1,35 @@
-# Compiled class file
-*.class
+# Eclipse
+.project
+.classpath
+.settings/
+bin/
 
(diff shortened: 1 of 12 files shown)


rsvoboda:
Downgrade to JUnit 4 to avoid class mode instead of suite mode for Arquillian

JUnit 5 confuses Arquillian somehow and makes him stop and start the server after each TestCase. This is not acceptable from execution time perspective.

PR #20, opened 2019-11-06
Add commom module from TT QE test suite with Docker utilities

mnovak1:


mnovak1:
@zhantemirov As we discussed today this is the promised docker tooling, I've added just the jaeger container as I need it in my test. I expect that other containers will be added per need.

Could you review it, please?

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+
+    private final ExecutorService outputPrinter = Executors.newSingleThreadExecutor();
+
+    public Docker(String name, String image) {
+        this.uuid = UUID.randomUUID().toString();
+        this.name = name;
+        this.image = image;
+    }
+
+    public Docker waitForLogLine(String logWait) {
```
honza-kasik: These methods are in fact setters, yet their name doesn't reflect this fact. From the name I would guess that calling this method causes some kind of wait. Also, JavaDoc is missing...

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+        this.awaitedLogLine = logWait;
+        return this;
+    }
+
+    public Docker waitTimeout(long timeoutInMillis) {
```
honza-kasik: Not expected method name, missing JavaDoc.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+        this.waitTimeoutInMillis = timeoutInMillis;
+        return this;
+    }
+
+    public Docker port(String port) {
+        this.ports.add(port);
+        return this;
+    }
+
+    public Docker envVar(String key, String value) {
```
honza-kasik: Not expected method name, missing JavaDoc.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+        this.awaitedLogLine = logWait;
+        return this;
+    }
+
+    public Docker waitTimeout(long timeoutInMillis) {
+        this.waitTimeoutInMillis = timeoutInMillis;
+        return this;
+    }
+
+    public Docker port(String port) {
```
honza-kasik: Not expected method name, missing JavaDoc.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+        this.ports.add(port);
+        return this;
+    }
+
+    public Docker envVar(String key, String value) {
+        environmentVariables.put(key, value);
+        return this;
+    }
+
+    public Docker cmdArg(String commandArgument) {
```
honza-kasik: Not expected method name, missing JavaDoc.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+        for (Map.Entry<String, String> envVar : environmentVariables.entrySet()) {
+            cmd.add("-e");
+            cmd.add(envVar.getKey() + "=" + envVar.getValue());
+        }
+
+        cmd.add(image);
+
+        cmd.addAll(commandArguments);
+
+        System.out.println(Ansi.ansi().reset().a("Starting container ").fgCyan().a(name).reset()
```
honza-kasik: Do we want to log using `System.out`?

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+        environmentVariables.put(key, value);
+        return this;
+    }
+
+    public Docker cmdArg(String commandArgument) {
+        commandArguments.add(commandArgument);
+        return this;
+    }
+
+    private void start() throws IOException, TimeoutException, InterruptedException {
```
honza-kasik: This method is unnecessary long.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+                // ignored because we're about to throw another exception
+            }
+
+            throw new TimeoutException("Container '" + name + " (" + uuid + ")' didn't print '"
+                    + awaitedLogLine + "' in " + waitTimeoutInMillis + " ms");
+        }
+    }
+
+    private void stop() throws IOException, InterruptedException {
+        System.out.println(Ansi.ansi().reset().a("Stopping container ").fgCyan().a(name).reset()
```
honza-kasik: Please use other logging method.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+    protected void before() throws Throwable {
+        start();
+    }
+
+    @Override
+    protected void after() {
+        try {
+            stop();
+        } catch (IOException | InterruptedException e) {
+            System.out.println(Ansi.ansi().reset().a("Failed stopping container ").fgCyan().a(name).reset()
```
honza-kasik: What about Java logging or Slf4j? Also, do we want to swallow the exception and continue if container is not stopped?

common/src/main/java/org/jboss/eap/qe/ts/common/docker/DockerContainers.java
```diff
@@ -0,0 +1,23 @@
+package org.jboss.eap.qe.ts.common.docker;
+
+public class DockerContainers {
```
honza-kasik: JavaDoc missing.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Intended to be used as a JUnit {@link org.junit.ClassRule ClassRule}.
```
honza-kasik: Example of usage is there which is great, but the basic description of what this class does is missing.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+                .command(cmd)
+                .start();
+
+        CountDownLatch latch = new CountDownLatch(1);
+
+        outputPrinter.execute(() -> {
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(dockerRunProcess.getInputStream()))) {
+                String line;
+                while ((line = reader.readLine()) != null) {
+                    System.out.println(Ansi.ansi().fgCyan().a(name).reset().a("> ").a(line));
```
honza-kasik: `System.out` again.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+
+                    if (awaitedLogLine != null && line.contains(awaitedLogLine)) {
+                        latch.countDown();
+                    }
+                }
+            } catch (IOException ignored) {
```
honza-kasik: It would be great to leave a comment why is the exception ignored.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,188 @@
...
+
+        if (awaitedLogLine != null) {
+            if (latch.await(waitTimeoutInMillis, TimeUnit.MILLISECONDS)) {
+                return;
+            }
+
+            try {
+                stop();
+            } catch (Exception ignored) {
+                // ignored because we're about to throw another exception
```
honza-kasik: But it could failed to stop and user is not informed about this fact.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
mnovak1: @honza-kasik thanks for the review.

Yes, this is copy-pasty from TT QE test suite. I didn't make any modification into API or logic.

In case of Builder pattern having `set...()` methods is not must have. It's more about to have descriptive names. For example `StringBuilder.append()`. 

One argument why to to keep current method names is that it will make migration easier and we will not have to rewrite it but looking at TT QE test suite this is basically used just in `DockerContainers` class so it's not big effort.

I'll change the names of methods to better reflect what is set. I would like to avoid writing JavaDoc for those methods as those should be self-explanatory from method names.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/DockerContainers.java
mnovak1: Makes sense. I will add description what is the purpose of this class.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
mnovak1: Correct, we'll need to add some logger which would be used in the whole test suite. Any preference/suggestion here? 

What about to use lombok `@Log4j` - https://projectlombok.org/features/log

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
honza-kasik: Thanks for considering this! Although this look like a `Builder` pattern it is IMHO not. There is no separation between the builder and product. So these are in my opinion setters and not methods of builder.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
honza-kasik: I didn't even know that Lombok project exists. Looks good - I'd like to use it in the testsuite. Personally, I prefer Slf4j, but that's just a matter of habit. Thank you for the tip!

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
mnovak1: Looking at the code this is just logging output from docker container. Once `Docker.stop()` is called at the end of the test then this should throw `IOException` as there is no graceful shutdown hook. I'll add a comment.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
mnovak1: I'll take closer look but as docker output is logged then any failure of docker container would be already logged. 

Also here is called `stop()` which is in fact just calling `docker stop ...` command using `ProcessBuilder`. I don't see any useful information which might go out of this method. Better would be to adjust `stop()` method to log output from `docker stop ...` command.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
mnovak1: @honza-kasik Don't worry I've changed the method names expect this one :-) Could you take a look, please?

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
mnovak1: I've added comment why it's ignored.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
mnovak1: Done.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
mnovak1: I have no problem with Slf4j as well. I'll bring this topic on chat with others.

mnovak1:
@rsvoboda thanks for pointing to this project, I was not aware of it and I'll use it in the test. With this in mind it does not make sense to create any common module with tooling for docker and use it directly in test. I'll close this PR.

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
istraka: What about having it in some sort of README file?

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,193 @@
...
+        this.ports.add(port);
+        return this;
+    }
+
+    public Docker addEnvVar(String key, String value) {
+        environmentVariables.put(key, value);
+        return this;
+    }
+
+    public Docker addCmdArg(String commandArgument) {
```
istraka: I see lots of setters. Are all of them mandatory? It should be documented which are mandatory. Or create some builder with proper javadoc (where we can see defaults and so on)

common/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -0,0 +1,193 @@
...
+        outputPrinter.shutdown();
+        outputPrinter.awaitTermination(10, TimeUnit.SECONDS);
+
+        new ProcessBuilder()
+                .command("docker", "rm", uuid)
+                .start()
+                .waitFor(10, TimeUnit.SECONDS);
+    }
+
+    // ---
```
istraka: Remove please.

mnovak1:
@bsideup hi, I saw you to be heavily involved into testcontainers. Thus asking you ;-)

I did some evaluation of https://github.com/testcontainers and figured out that it does not work with Podman on RHEL 8 (which is one of supported/tested platforms for Wildfly/EAP) and requires docker daemon to be present. 

Is there a way or plan to add this functionality (work with podman) into testcontainers?

bsideup:
@mnovak1 

Hi! Testcontainers uses Docker's REST API. I am not sure if there is something like that for podman (since it is only a subset of Docker's functionality)
Currently we do not have any concrete plans to support Podman itself (since we would need to write a daemon or some proxy for it and expose as REST API, for Linux, Mac and Windows), and we're more excited about the rootless Docker efforts

PR #25, opened 2019-11-14
Travis CI

rsvoboda:
Travis CI to ensure compilation works fine

Test execution against WF master of WF feature branch is not trivial.
Could be solved by custom build script of WF master (in the future).


rsvoboda:
@mnovak1 fyi

PR #30, opened 2019-11-15
Refactoring of setContainerReadyTimeout, Jaeger ReadyCondition simplification

rsvoboda:
Refactoring of setContainerReadyTimeout
Jaeger ReadyCondition simplification


tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -134,10 +134,11 @@ public Builder(String name, String image) {
         /**
          * Timeout in millis to wait until container is ready/starts
```
mnovak1: `in millis` should be removed as well.

rsvoboda: ok, will do in next fixes bulk

PR #31, opened 2019-11-18
Provide fail-fast mechanism for starting docker container - fixes htt…

mnovak1:
…ps://github.com/jboss-eap-qe/eap-microprofile-test-suite/issues/28

### DIFF
--- a/tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
+++ b/tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
@@ -11,9 +11,11 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 /**
  * Utility class for starting docker containers. This class allows to start any docker container.
@@ -31,11 +33,15 @@ public class Docker extends ExternalResource {
     private ContainerReadyCondition containerReadyCondition;
     private long containerReadyTimeout;
     private final ExecutorService outputPrinter = Executors.newSingleThreadExecutor();
+    private Process dockerRunProcess;
 
     private Docker() {
     } // avoid instantiation, use Builder
 
     protected void start() throws Exception {
+
+        checkDockerPresent();
+
         List<String> cmd = new ArrayList<>();
 
         cmd.add("docker");
@@ -60,7 +66,7 @@ public class Docker extends ExternalResource {
         System.out.println(Ansi.ansi().reset().a("Starting container ").fgCyan().a(name).reset()
                 .a(" with ID ").fgYellow().a(uuid).reset());
 
-        Process dockerRunProcess = new ProcessBuilder()
+        dockerRunProcess = new ProcessBuilder()
                 .redirectErrorStream(true)
                 .command(cmd)
                 .start();
@@ -77,7 +83,70 @@ public class Docker extends ExternalResource {
             }
         });
 
-        containerReadyCondition.waitUntilReady(containerReadyTimeout);
+        long startTime = System.currentTimeMillis();
+        while (!isContainerReady()) {
+            if (System.currentTimeMillis() - startTime > containerReadyTimeout) {
+                stop();
+                throw new DockerTimeoutException(uuid + "- Container was not ready in timeout : " + containerReadyTimeout + " ms");
+            }
+            // fail fast mechanism in case of malformed docker command, for example bad arguments, invalid format of port mapping, image version,...
+            if (!dockerRunProcess.isAlive() && dockerRunProcess.exitValue() != 0) {
+                throw new DockerException(uuid + "- Starting of docker container using command: \"" + String.join(" ", cmd)
+                        + "\" failed. Check that provided command is correct.");
+            }
+        }
+    }
+
+    private boolean isContainerReady() throws Exception {
+        CompletableFuture<Boolean> condition = new CompletableFuture().supplyAsync(() -> containerReadyCondition.isReady());
+        try {
+            return condition.get(containerReadyTimeout, TimeUnit.MILLISECONDS);
+        } catch (TimeoutException ex) {
+            stop();
+            // in case condition hangs interrupt it so there are no zombie threads
+            condition.cancel(true);
+
+            throw new ContainerReadyConditionException(uuid + "- Provided ContainerReadyCondition.isReady() method took longer than containerReadyTimeout: "
+                    + containerReadyTimeout + " ms. Check it does not hang and does not take longer then containerReadyTimeout. " +
+                    "It's expected that ContainerReadyCondition.isReady() method is short lived (takes less than 1 second).", ex);
+        }
+    }
+
+    private void checkDockerPresent() throws Exception {
+        Process dockerInfoProcess = new ProcessBuilder()
+                .redirectErrorStream(true)
+                .command(new String[] {"docker", "info"})
+                .start();
+        dockerInfoProcess.waitFor();
+        if (dockerInfoProcess.exitValue() != 0)   {
+            throw new DockerException("Docker is either not present or not installed on this machine. It must be installed " +
+                    "and started up for executing tests with docker container.");
+        }
+    }
+
+    /**
+     * @return Returns true if docker container is running. It does NOT check whether container is ready.
+     */
+    protected boolean isRunning() throws Exception {
+        Process dockerRunProcess = new ProcessBuilder()
+                .redirectErrorStream(true)
+                .command(new String[] {"docker", "ps"})
+                .start();
+
+        dockerRunProcess.waitFor();
+
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(dockerRunProcess.getInputStream()))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                if (line.contains(uuid))    {
+                    return true;
+                }
+            }
+        } catch (IOException ignored) {
+            // ignore as any stop of docker container breaks the reader stream
+            // note that shutdown of docker would be already logged
+        }
+        return false;
     }
 
     protected void stop() throws Exception {
@@ -126,7 +195,7 @@ public class Docker extends ExternalResource {
         private ContainerReadyCondition containerReadyCondition = () -> true;
 
         public Builder(String name, String image) {
-            this.uuid = UUID.randomUUID().toString();
+            this.uuid = name + "-" + UUID.randomUUID().toString();
             this.name = name;
             this.image = image;
         }
@@ -135,7 +204,7 @@ public class Docker extends ExternalResource {
          * Timeout to wait until container is ready/starts
          *
          * @param timeout the maximum time to wait
-         * @param unit the time unit of the {@code timeout} argument
+         * @param unit    the time unit of the {@code timeout} argument
          */
         public Builder setContainerReadyTimeout(long timeout, TimeUnit unit) {
             this.containerReadyTimeoutInMillis = unit.toMillis(timeout);
--- a/tooling-docker/src/test/java/org/jboss/eap/qe/ts/common/docker/MalformedDockerConfigTest.java
+++ b/tooling-docker/src/test/java/org/jboss/eap/qe/ts/common/docker/MalformedDockerConfigTest.java
@@ -1,0 +1,94 @@
+package org.jboss.eap.qe.ts.common.docker;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.concurrent.TimeUnit;
+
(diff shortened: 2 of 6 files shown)


tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -77,7 +78,17 @@ protected void start() throws Exception {
             }
         });
 
-        containerReadyCondition.waitUntilReady(containerReadyTimeout);
+        long startTime = System.currentTimeMillis();
+        while (!(new CompletableFuture().supplyAsync(() -> containerReadyCondition.isReady())).get()) {
```
honza-kasik: Just a curious question: Why is this async when there is a waiting for result?

mnovak1: I did some tries to specify timeout on isReady() method so in case it's badly implemented, it does not hang forever and `containerReadyTimeout` will be applied however I did not finish it. I've changed the code so it's applied and added test for it.

Now when containerReadyCondition.isReady() will hang then after `containerReadyTimeout` new exception will be thrown.

mnovak1: I'll mark this as resolved.

mnovak1:
@rsvoboda thanks for spotting this, fixed and re-pushed.

mnovak1:
@honza-kasik @zhantemirov @fabiobrz hi guys, could you take a look as well, please? thanks

tooling-docker/src/test/java/org/jboss/eap/qe/ts/common/docker/MalformedDockerConfigTest.java
```diff
@@ -0,0 +1,57 @@
...
+        thrown.expect(Exception.class);
+        thrown.expectMessage(startsWith("Starting of docker container using command: \"docker run --name"));
+        thrown.expectMessage(endsWith("failed. Check that provided command is correct."));
+
+        containerWithInvalidVersion.start();
+    }
+
+    @Test
+    public void testHangingReadyCondition() throws Exception {
+        Docker containerWithInvalidVersion = new Docker.Builder("wildfly", "jboss/wildfly:18.0.0.Final")
```
fabiobrz: I would change this from `containerWithInvalidVersion` to `hangingForReadyConditionContainer` or something similar.

tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -77,7 +79,31 @@ protected void start() throws Exception {
...
         });
 
-        containerReadyCondition.waitUntilReady(containerReadyTimeout);
+        long startTime = System.currentTimeMillis();
+        while (!isContainerReady()) {
+            if (System.currentTimeMillis() - startTime > containerReadyTimeout) {
+                throw new Exception("Container was not ready in timeout : " + containerReadyTimeout + " ms");
+            }
+            // fail fast mechanism in case of malformed docker command, for example bad arguments, invalid format of port mapping, image version,...
+            if (!dockerRunProcess.isAlive() && dockerRunProcess.exitValue() != 0) {
```
fabiobrz: Either one earlier check is to be done on docker process alone or I think I am getting a couple of green tests because of this - in conjunction with expected exception prefix [and suffix] as defined in test classes - even if I have docker daemon not running. 
WDYT?

tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -77,7 +79,31 @@ protected void start() throws Exception {
...
+
+    private boolean isContainerReady() throws Exception {
+        CompletableFuture<Boolean> condition = new CompletableFuture().supplyAsync(() -> containerReadyCondition.isReady());
+        try {
+            return condition.get(containerReadyTimeout, TimeUnit.MILLISECONDS);
+        } catch (TimeoutException ex) {
+            // in case condition hangs interrupt it so there are no zombie threads
+            condition.cancel(true);
+
+            throw new Exception("Provided ContainerReadyCondition.isReady() method took longer than containerReadyTimeout: "
```
fabiobrz: Could be providing a couple of strongly typed `Exception` sub-classes a good idea for refactoring? WDYT?

tooling-docker/src/test/java/org/jboss/eap/qe/ts/common/docker/MalformedDockerConfigTest.java
mnovak1: what about `testContainerReadyTimeoutWithHangingCondition`?

tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
mnovak1: I'm not sure I understand. Do you want to update exception message that docker daemon might not be running?

tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
mnovak1: +1 I was too lazy to do so :-) I'll fix it.

tooling-docker/src/test/java/org/jboss/eap/qe/ts/common/docker/MalformedDockerConfigTest.java
fabiobrz: Hi, it looks like the name for a _condition_, actually, rather than for something that represents a Docker instance. Using the same principle that you followed in the previous method (`containerWith*`) I'd have gone with something similar, like `containerWithHangingReadyCondition`. That said, I wouldn't want for us to trip over variable names too much, here.

mnovak1: `testContainerWithHangingReadyCondition` it is ;-)

tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
fabiobrz: It is not just about the exception message.
The message is the same - i.e. the expected one - that the `testFailFastWithMalformedDockerCommand` method is supposing to catch. 
So my concern is mainly about not getting a green bar from a fail-safe case method related to an unavailable image when instead tests have not been executed, actually - which instead I happen to be able to do.
That said I thought about a pre execution check for docker process to be running. WDYT?

mnovak1: Thanks, now i get your point. I'll just rephrase it, we'll add check that docker command is present. This is because there is podman on rhel8 which does not have any docker daemon. 

I've added the check and re-pushed.

tooling-docker/src/test/java/org/jboss/eap/qe/ts/common/docker/MalformedDockerConfigTest.java
fabiobrz: Now I get where we clashed... I meant the `Docker` variable name, while you addressed the method name. Anyway, if you don't want to change it, that's fine to me too.

tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -77,7 +82,42 @@ protected void start() throws Exception {
...
+    }
+
+    private void checkDockerPresent() throws Exception {
+        Process dockerRunProcess = new ProcessBuilder()
+                .redirectErrorStream(true)
+                .command(new String[] {"docker", "info"})
+                .start();
+        dockerRunProcess.waitFor();
+        if (dockerRunProcess.exitValue() != 0)   {
+            throw new DockerException("Docker is not present on this machine. It must be installed for running tests with docker container.");
```
fabiobrz: I'd rephrase 
`Docker is not present on this machine. It must be installed for running tests with docker container.`
to
`Docker is either not present or not installed on this machine. It must be installed and started up for executing tests with docker container.`

mnovak1: thanks, fixed

rsvoboda:
I have tried this on my home/slow machine with Windows and Docker Toolbox.
My machine is too slow to start 2 WF instances in time, what a great opportunity to see problems.

**1) error message**
```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 121.011 s <<< FAILURE! - in org.jboss.eap.qe.ts.common.docker.DockerTest
[ERROR] org.jboss.eap.qe.ts.common.docker.DockerTest  Time elapsed: 121.008 s  <<< ERROR!
org.jboss.eap.qe.ts.common.docker.DockerTimeoutException: Container was not ready in timeout : 120000 ms
        at org.jboss.eap.qe.ts.common.docker.Docker.start(Docker.java:88)
        at org.jboss.eap.qe.ts.common.docker.Docker.before(Docker.java:144)
```
Error message needs to mention container name for easy identification (was it wildfly-server-one or wildfly-server-two ?)
Container __TODO__ADD_NAME was not ready in timeout : 120000 ms


**2) running process for first WF after TS is finished**
```
$ docker ps
CONTAINER ID        IMAGE                        COMMAND                  CREATED             STATUS              PORTS                                              NAMES
220f0b63f5f6        jboss/wildfly:18.0.0.Final   "/opt/jboss/wildfly/"   5 minutes ago       Up 5 minutes        0.0.0.0:22222->8080/tcp, 0.0.0.0:22990->9990/tcp   1efb1c30-9130-496b-9e04-30107b4d85c9
```
Better cleanup needs to happen when the TestCase finishes.

In Quarkus QS there is stop command called before start to eliminate conflict with a leftover. I prefer to clean stuff right after the tests stops/fails, calling stop before running the container is workaround which could be applied if the "proper" way hits some limits
https://github.com/quarkusio/quarkus-quickstarts/blob/master/mqtt-quickstart/pom.xml#L97-L98

An option to skip such final cleanup for investigation can be handy too, could be tracked as a new issue.


**3) better names for started container**
1efb1c30-9130-496b-9e04-30107b4d85c9 is terrible thing for human investigation
We should have a human-readable name or at least prefix, at least something like wildfly-server-two-1efb1c30-9130-496b-9e04-30107b4d85c9 or something more stable / "rule base" like ${TEST_NAME}-wildfly-server-two
Could be tracked as a new issue.

Didn't look at code yet ;)

tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -77,7 +82,43 @@ protected void start() throws Exception {
...
+            // in case condition hangs interrupt it so there are no zombie threads
+            condition.cancel(true);
+
+            throw new ContainerReadyConditionException("Provided ContainerReadyCondition.isReady() method took longer than containerReadyTimeout: "
+                    + containerReadyTimeout + " ms. Check it does not hang and does not take longer then containerReadyTimeout. " +
+                    "It's expected that ContainerReadyCondition.isReady() method is short lived (takes less than 1 second).", ex);
+        }
+    }
+
+    private void checkDockerPresent() throws Exception {
```
honza-kasik: Shouldn't we check the output too? There might be `alias docker=podman` set up on the system (`podman-docker` package [does this for example when installed](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/building_running_and_managing_containers/index#running-containers-without-docker_building-running-and-managing-containers)) and if we are looking specifically for the Docker, this might be an issue.

mnovak1: @honza-kasik thanks, this check works with docker and podman as well. I've tried it on RHEL8.

mnovak1:
@rsvoboda I've fixed 1) and 3). I'll take a look at 2) later today.

tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
honza-kasik: Thanks, that is not my concern. The issue is: Do we want to proceed with Podman?

mnovak1: Podman is pushed on RHEL8 and at least for this moment we'll align with this. Current tooling depends on basic usage of `docker` command only so we're ok with having `alias docker=podman`.

tooling-docker/src/test/java/org/jboss/eap/qe/ts/common/docker/MalformedDockerConfigTest.java
mnovak1: Right :-), I've changed it.

mnovak1:
I've fixed 2) as well and added test for container ready timeout. I've also enhanced some tests with assert there is no left docker container.

mnovak1:
All the above mentioned issues are fixed. Tests passed on RHEL7 and RHEL8. 

I would like to merge this as it has grown too big. In case there will be found new issues they'll be handled in separate PR.

PR #32, opened 2019-11-18
TS cleanup and small enhancements

rsvoboda:
- Removal of redundant RestAssured.when()
 - Typo fix for Docker method name and JavaDoc
 - Removal of commented out text in pom.xml


rsvoboda:
We should get rid of `import static org.hamcrest.Matchers.*;` too, will be part of separate PR once this gets merged

tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -148,7 +148,7 @@ public Builder setContainerReadyTimeout(long timeout, TimeUnit unit) {
          *
          * @param portMapping port mapping as defined by docker command, for exampple "8080:80" or "8080:80/tcp"
          */
-        public Builder wihtPortMapping(String portMapping) {
+        public Builder withPortMapping(String portMapping) {
```
mnovak1: Nice catch, thanks :-)

mnovak1:
LGTM

PR #40, opened 2019-11-21
Introduce dependencyManagement in root pom.xml

rsvoboda:
Introduce dependencyManagement in root pom.xml

This is consequence of review of https://github.com/jboss-eap-qe/eap-microprofile-test-suite/pull/38

We need this asap, adding you all as reviewers to be aware. One ack is sufficient to merge.


microprofile-metrics/pom.xml
```diff
@@ -9,8 +9,33 @@
...
     <artifactId>microprofile-metrics</artifactId>
-    
-    <dependencies>
 
+    <dependencies>
+        <dependency>
+            <groupId>org.eclipse.microprofile</groupId>
+            <artifactId>microprofile</artifactId>
+            <type>pom</type>
+            <scope>provided</scope>
```
mnovak1: We shouldn't duplicate specifying scope in `dependencyManagement` and then in `dependencies` in each submodule. I've tried to google some examples and best practices but did not find definitive answer. 

But based on discussion in stackoverflow[1] i'm more for defining scope in submodules. wdyt?

[1] https://stackoverflow.com/questions/15221299/dependencymanagement-and-scope

rsvoboda: I added scopes to root pom.xml intentionally

 - people will probably copy those deps to their new module, throw away just version line, they won't think about defining scope
 - previously I did some deps analysis based on dependencyManagement definition and missing scope definition was big trouble for that analysis so I tend to keep scope there too

PR #74, opened 2019-12-05
Added solution for watching logs using management model

honza-kasik:
Added solution for watching logs using management model. Do you like this feature?

- [x] Pull Request contains description for the changes
- [x] Pull Request does not include fixes for multiple issues / topics
- [x] Code compiles and tests are passing


tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/LogWatcher.java
```diff
@@ -0,0 +1,17 @@
+package org.jboss.eap.qe.microprofile.tooling.server.logwatch;
+
+import java.util.regex.Pattern;
+
+/**
+ * Interface describing log watcher which checks log for a pattern.
+ */
+public interface LogWatcher {
```
istraka: `Watcher` inclines it watch the logs permanently. What about `LogReader`?

tooling-server-configuration/src/test/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/ModelNodeLogWatcherTestCase.java
```diff
@@ -0,0 +1,67 @@
...
+    public void testLineNoMatchedClient() throws ConfigurationException, IOException {
+        try (final OnlineManagementClient client = ManagementClientProvider.onlineStandalone()) {
+            final LogWatcher logWatcher = new ModelNodeLogWatcher(client, 10, true);
+
+            Assert.assertFalse(logWatcher.wasLineWithPatternLogged(Pattern.compile(".*Foooqux 42.*")));
+        }
+    }
+
+    @Test
+    @RunAsClient
```
istraka: The test is not supposed to run as a client I suppose.

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/ModelNodeLogWatcher.java
```diff
@@ -0,0 +1,74 @@
+package org.jboss.eap.qe.microprofile.tooling.server.logwatch;
+
+import org.jboss.dmr.ModelNode;
+import org.wildfly.extras.creaper.core.online.ModelNodeResult;
```
istraka: @mnovak1 you have run into some issues using Creaper with in-container Arquillian mode. However there are tests to be executed in such mode. Any thoughts?

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/LogWatcher.java
```diff
@@ -0,0 +1,17 @@
...
+
+    /**
+     * Perform search in log or its excerpt and find if a line matches the pattern.
+     * @param pattern a pattern which will be the log line matched against
+     * @return true if log line matching pattern was found in log, false otherwise
+     */
+    boolean wasLineWithPatternLogged(final Pattern pattern);
```
istraka: Can it be shorted? Something like `isLogged`?

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/ModelNodeLogWatcher.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.io.IOException;
+import java.util.Optional;
+import java.util.regex.Pattern;
+
+/**
+ * Log Watcher based on accessing messages through {@link org.jboss.dmr.ModelNode} returned from {@code :read-log-file}
+ * operation.
+ */
+public final class ModelNodeLogWatcher implements LogWatcher {
```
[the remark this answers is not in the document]
honza-kasik: I am against the idea of adding such constructor. As it was said before - logs can get pretty big and you don't want to get all that in the memory. The only option is to have constructor which would read some default amount of lines.

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/ModelNodeLogWatcher.java
mnovak1: @istraka in case it will be in-container test then the whole creaper core (with wildfly-cli and wildfly-controller-client must be added)

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/LogWatcher.java
honza-kasik: It can, but IMHO it is great addition to have such self descriptive method name (that it matches lines against pattern).

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/ModelNodeLogWatcher.java
honza-kasik: yep, we will need to have a deployment which would contain all dependencies necessary. So I am thinking about removing all non-client tests.

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/ModelNodeLogWatcher.java
```diff
@@ -0,0 +1,74 @@
...
+                    Values.of("lines", this.countOfLines)
+                            .and("tail", this.readLogFromEnd));
+
+            result.assertSuccess("Reading log file failed!");
+
+            final Optional<String> optionalMatch = result.value()
+                    .asList()
+                    .stream()
+                    .map(ModelNode::asString)
+                    .filter((final String line) -> pattern.matcher(line).matches())
```
istraka: Using this it would check all the lines. We want to find just the first one.

honza-kasik: It will check all lines read from the file amount of lines is equal to value passed to constructor). Is this something we don't want to do?

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/ModelNodeLogWatcher.java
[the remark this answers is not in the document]
istraka: OK. Can you add to the JavaDoc `-1` option which should get all the lines? And a test? 

I see what you are saying, however, IMHO it is not a good idea to limit developers. This might encourage them to think about it.

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/LogWatcher.java
istraka: What about `wasLineLogged` then? Parameter is also descriptive. Moreover can you overload method with `String` parameter? It might come handy to just past `.*WFLY-XYZ.*` string.

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/ModelNodeLogWatcher.java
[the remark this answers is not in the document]
honza-kasik: +1 - I'll add it. In my opinion, some restriction makes sense. User (developer) must be aware of fact, that some amount of lines is being read into the memory which can have consequences. Mindlessly calling constructor with least amount of parameters could lead to some memory issues...

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/ModelNodeLogWatcher.java
istraka: If we are checking 100k lines and the one is in the middle 50% of the work is pointless - line was logged.

honza-kasik: Oh, thanks! I see what you are thinking of now :) I'll think of it.

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/ModelNodeLogWatcher.java
fabiobrz: > in case it will be in-container test then the whole creaper core (with wildfly-cli and wildfly-controller-client must be added)

@mnovak1, @honza-kasik could a test for in-container usage be added once this gets merged?

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/ModelNodeLogWatcher.java
honza-kasik: I thought about it and it should be OK as it is. Streams are lazy, so all lines won't be evaluated. The search stop when any line matches.

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/LogWatcher.java
honza-kasik: +1 for name change. I think that you will lost the readability and IDE regex support if you place just a `String` there when you really want to pass a pattern.

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/ModelNodeLogWatcher.java
honza-kasik: @fabiobrz I think that it would be the best solution, when @mnovak1 has actually experience doing this.

mnovak1: @fabiobrz I'm not sure when I get to it. Maybe tomorrow so anyone is free to try :-)

honza-kasik:
Made suggested changes, ready to be merged.

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/LogWatcher.java
istraka: Agree. However I want two methods. `wasLineLogged(Pattern pattern)` and `wasLineLogged(String pattern)`. So we can decide if we want to always write `assert logReader.wasLineLogged(Pattern.compile("*.message.*);` or just `assert logReader.wasLineLogged(".*message.*)l`

honza-kasik: I know what you want. It just IMHO doesn't make a sense. Either you want to pass a `String` - in that case I suggest it will be automatically considered to be desired sub string in line which will be looked after (o you will actually call `logReader.wasLineLogged("message");`) or you want to pass a pattern in which case you will call `logReader.wasLineLogged(Pattern.compile("*.message.*);`. I don't want a masquerade where you are passing a `String` but it is actually a pattern... Are you OK with this?

istraka: I see. Good point with a sub string. I just wanted to get rid of  `Pattern.compile`. Something like `String::matches` that would save some typing. But I am OK with what we have now,

honza-kasik:
@istraka See latest commit.

istraka:
LGTM

honza-kasik:
@rsvoboda Are you OK with this approach?

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/LogReader.java
```diff
@@ -0,0 +1,24 @@
+package org.jboss.eap.qe.microprofile.tooling.server.logwatch;
+
+import java.util.regex.Pattern;
+
+/**
+ * Interface describing log watcher which checks log for a pattern.
```
rsvoboda: watcher ?

On multiple places

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/LogReader.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+public interface LogReader {
+
+    /**
+     * Perform search in log or its excerpt and find if a line that matches the pattern.
+     * @param pattern a pattern which will be the log line matched against
+     * @return true if log line matching pattern was found in log, false otherwise
+     */
+    boolean wasLineLogged(final Pattern pattern);
```
rsvoboda: Do not like method name (LogReader.wasLineLogged()) too much. Contains / matches sound like better keywords.

What about simple `Log.contains()`

tooling-server-configuration/src/main/java/org/jboss/eap/qe/microprofile/tooling/server/logwatch/LogReader.java
```diff
@@ -0,0 +1,24 @@
+package org.jboss.eap.qe.microprofile.tooling.server.logwatch;
```
rsvoboda: logwatch => log

rsvoboda:
```java
ServerLog serverLog = new ModelNodeServerLog(client, 50);
...
if ( serverLog.contains("WFLY00035") ) {
    ...
}
```
@istraka @mnovak1 @honza-kasik @fabiobrz  what about this way ?

I was discussing with Jan some changes, for now we agreed on this: `LogChecker.logContains()`

honza-kasik:
@rsvoboda All comments resolved

rsvoboda:
I'm still in favor of ServerLog.contains but I will merge it and create issue to switch to ServerLog.contains (maybe / one day ...) We can discuss it more on Wednesday, code is in one repo, refactoring is not hard at all.

I want to merge it so we can have test cases with log inspection landing in the repo.

Issue #46, opened 2019-11-25
Update PR template to include TPG stuff


rsvoboda:
Update PR template to include TPG stuff

details in https://github.com/jboss-eap-qe/eap-microprofile-test-suite/issues/44

rsvoboda:
```
@tpTestDetails __TODO__
@tpPassCrit __TODO__
@tpSince EAP 7.4.0.CD19
```

rsvoboda:
Template updated - https://github.com/jboss-eap-qe/eap-microprofile-test-suite/pull/49

PR #110, opened 2019-12-17
Added test veryfing all key sizes are supported

honza-kasik:
- N/A Link to the passing job is provided (No job can currently handle the subsequent build of projects needed)


microprofile-jwt/src/test/java/org/jboss/eap/qe/microprofile/jwt/security/publickeylocation/PublicKeyFileTestCase.java
```diff
@@ -0,0 +1,4 @@
+package org.jboss.eap.qe.microprofile.jwt.security.publickeylocation;
+
+public class PublicKeyFileTestCase {
```
fabiobrz: I guess this one is empty because it will be needed for next iteration, right? Just to be sure.

microprofile-jwt/src/test/resources/pki/README.md
```diff
@@ -1,15 +1,15 @@
...
-## How to recreate
-```bash
-ssh-keygen -t rsa -b 2048 -m PEM -f key.private.pem #with empty passphrase!
-openssl pkcs8 -topk8 -v2 des3 -nocrypt \
-    -in key.private.pem \
-    -out key.private.pkcs8.pem
-openssl rsa -in key.private.pkcs8.pem -pubout -outform PEM -out key.public.pem
-rm key.private.pem key.private.pem.pub #cleanup
-```
+* `key.private.pkcs8.pem` and `key.public.pem` - this pair is the main key pair use in most tests, 2048 bits long
```
fabiobrz: I am wondering if some kind of prefix - like `main` or better (IMO) `default` wouldn't be good anyway. WDYT?

microprofile-jwt/src/test/java/org/jboss/eap/qe/microprofile/jwt/security/publickeylocation/PublicKeyFileTestCase.java
honza-kasik: Yep, sorry, it shouldn't be included in this commit, will remove. Thank you!

microprofile-jwt/src/test/resources/pki/README.md
honza-kasik: I was wondering about this too. I think that if it would have "default" in its name someone might assume that it is chosen purely based on this... I can add this prefix/suffix if you want.

honza-kasik:
Thanks @fabiobrz! I forced push commit with removed empty class...

microprofile-jwt/src/test/resources/pki/README.md
fabiobrz: Well, I think that what you just stated makes sense, so from my side no need for adding it.

microprofile-jwt/src/test/java/org/jboss/eap/qe/microprofile/jwt/security/keyproperties/KeySizeTestCase.java
```diff
@@ -0,0 +1,158 @@
...
+
+        RestAssured.given()
+                .header("Authorization", "Bearer " + token.getRawValue())
+                .when().get(url.toExternalForm() + "basic-endpoint")
+                .then()
+                .body(equalTo("<html><head><title>Error</title></head><body>Unauthorized</body></html>"))
+                .and()
+                .statusCode(401);
+
+        //TODO add check for expected log message with correct reason
```
fabiobrz: What is the `TODO` related to?

microprofile-jwt/src/test/resources/pki/README.md
fabiobrz: Yes, you reasoning makes sense and you also mentioned the fact in README, so from my side there's no need to add it.

microprofile-jwt/src/test/java/org/jboss/eap/qe/microprofile/jwt/security/keyproperties/KeySizeTestCase.java
honza-kasik: Thanks, removed for now, since the situation about JWT logs is a little bit unclear for now...

rsvoboda:
@fabiobrz can you please hit "Resolve conversation" buttons ?

fabiobrz:
> @fabiobrz can you please hit "Resolve conversation" buttons ?

Hi @rsvoboda, I can't actually see the above mentioned button, maybe I don't have the needed grants, but that's just my guess.

PR #124, opened 2019-12-20
MP OpenAPI - Adding security tests

fabiobrz:
Security scenario is covered to verify that endpoint can be accessed via HTTPS listener and a WARN is logged when HTTP is disabled.

- N/A Link to the passing job is provided: No job can currently handle the subsequent build of projects needed


microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/security/ListenerSecurityConfigurationTest.java
```diff
@@ -0,0 +1,239 @@
...
+     * @tpTestDetails Test to verify that missing activating MP OpenAPI when the HTTP (default) listener is disabled
+     *                causes the server to log a warning about MicroProfile OpenAPI specification requiring for the endpoint to
+     *                be accessible via [http]
+     * @tpPassCrit The server is logging expected {@code WFLYMPOAI0006} message because HTTP (default) listener
+     *             does not exist
+     *
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testWarningIsLoggedBecauseOfSkippedBadlyConfiguredDeployment() {
```
rsvoboda: where / how is the deployment "badly" configured ?

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/security/ListenerSecurityConfigurationTest.java
```diff
@@ -0,0 +1,239 @@
...
+     * @tpTestDetails Test to verify that the {@code openapi} endpoint is available via HTTPS once the listener
+     *                has been properly configured
+     * @tpPassCrit The {@code openapi} endpoint can be reached via [https] schema at related standard port number
+     *             (8443), response status is HTTP 200 and response body is not empty
+     *
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testOpenApiOnHTTPSListener() {
+        io.restassured.RestAssured.useRelaxedHTTPSValidation();
```
rsvoboda: is this relaxed https validation reset anywhere ?

if not then this test can cause side effect == all https checks will be relaxed
Atm we can live with it, but it must be at least tracked in issues

fabiobrz: No, definitely not resetting it. You're right, fixing...

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/security/ListenerSecurityConfigurationTest.java
fabiobrz: This is definitely a typo due to some quick c&p, sorry about that. going to fix it right now.

fabiobrz:
Hi @rsvoboda, addressed your comments, thanks.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/security/ListenerSecurityConfigurationTest.java
```diff
@@ -0,0 +1,238 @@
...
+     * @tpTestDetails Test to verify that activating MP OpenAPI when the HTTP (default) listener is disabled
+     *                causes the server to log a warning about MicroProfile OpenAPI specification requiring for the
+     *                endpoint to be accessible via [http]
+     * @tpPassCrit The server is logging expected {@code WFLYMPOAI0006} message because HTTP (default) listener
+     *             does not exist
+     *
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testWarningIsLoggedBecauseOfMissingDefaultHTTPListener() {
```
zhantemirov: Though the method name clearly states what is does, could it be shortened somehow? For instance, `MissingHTTPListenerWarningLogged` or something similar? WDYT?

fabiobrz: Yes, sounds good to me. I'll use what you suggested leaving the `test` prefix in place. Fixing...

fabiobrz:
Hi @zhantemirov I pushed the change that you suggested. Thanks!

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/OpenApiDeploymentUrlProvider.java
```diff
@@ -29,4 +29,16 @@ public static String composeDefaultOpenApiUrl() {
                 OpenApiTestConstants.DEFAULT_HOST_NAME,
                 OpenApiTestConstants.DEFAULT_ENDPOINT_PORT);
     }
+
+    /**
+     * Builds OpenAPI endpoint URL for HTTPS access
+     *
+     * @return
+     */
+    public static String composeHttpsOpenApiUrl() {
```
rsvoboda: This isn't needed/wanted, you can obtain it from ARQ

deployment url - https://docs.jboss.org/author/display/ARQ/ArquillianResource+injection
```java
@ArquillianResource
private URL baseURL;
```

management interface
```java
@ContainerResource
ManagementClient managementClient;

managementClient.getMgmtAddress() + ":" + managementClient.getMgmtPort()
```
https://github.com/wildfly/wildfly/blob/master/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/base/MicroProfileMetricsBaseTestCase.java#L57
https://github.com/wildfly/wildfly/blob/master/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/MetricsHelper.java#L34
https://github.com/wildfly/wildfly-arquillian/blob/master/common/src/main/java/org/jboss/as/arquillian/container/ManagementClient.java

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/security/ListenerSecurityConfigurationTest.java
```diff
@@ -0,0 +1,238 @@
...
+        return deployment;
+    }
+
+    static class OpenApiExtensionSetup implements ServerSetupTask {
+
+        @Override
+        public void setup(ManagementClient managementClient, String containerId) throws Exception {
+            httpsOpenApiUrl = OpenApiDeploymentUrlProvider.composeHttpsOpenApiUrl();
+            //  MP OpenAPI up
+            onlineManagementClient = ManagementClientProvider.onlineStandalone();
```
rsvoboda: from online management client you can get https port, in CLI it is `/socket-binding-group=standard-sockets/socket-binding=https/:read-attribute(name=bound-port)

But usually if ports are not standard they are shifted using `-Djboss.socket.binding.port-offset=` 
So from `@ArquillianResource private URL baseURL;` you can easily get https port number - `baseURL.getPort() - 8080 + 8443`

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/OpenApiTestConstants.java
```diff
@@ -6,4 +6,5 @@
 public class OpenApiTestConstants {
```
rsvoboda: These hard-coded constants needs to go away, see other comments about other ways

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/OpenApiDeploymentUrlProvider.java
fabiobrz: Thanks for the pointers, fixing.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/security/ListenerSecurityConfigurationTest.java
fabiobrz: Ok, I didn't consider offsets in general, so I think this would lead to a minor refactoring affecting already implemented tests, too.
I'll see how to provide minimal changes in separate commit, thanks for the explanation and hints.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/OpenApiTestConstants.java
fabiobrz: Yes, agree, fixing. Will provide separate commit in this PR following this change.

fabiobrz:
Hi @rsvoboda, thanks for you detailed input. 
Modifying the code following your suggestions led to some refactoring, mainly due to usage of injected (i.e. _non-static_) `@ArquillianResource`. 
So, `enableHTTPSListener`, `disableHTTPSListener`, `enableHTTPListener`, `disableHTTPSListener` moved from `ServerSetupTask` implementation into `testOpenApiOnHTTPSListener()` because otherwise - once that HTTP is disabled - Arquillian won't be able to inject the bound `HTTP` port ("undefined") and would fallback to 8080, which would definitely break everything when using `-Djboss.socket.binding.port-offset` with a non-zero value.

Because of this I had to give the tests a proper ordering through `@InSequence` annotation as the log check would fail when executed _before_ the HTTPS tests as no WARN is logged until HTTPS is enabled and `reload` performed. 

Also - I've been fixing the path for `keystore.jks` - using `ArquillianDescriptor` as source for current `jboss` named container home, as it was failing when executing tests `in-container` mode.

I also moved code from @BeforeClass and @AfterClass methods into ServerSetuTask implementation, as this is the most appropriate way to configure server.

Now, onto the `port offset` topic. This was not taken into account in _any_ of the MP OpenAPI module tests, so I provided a separate commit to fix this through a minimal - but I think needed - refactoring. 
Let's keep in mind that `arquillian.xml` must be properly modified when using `-Djboss.socket.binding.port-offset` with a non-zero value, here's a snippet I used during my tests:

```
<?xml version="1.0" encoding="UTF-8"?>
<arquillian xmlns="http://jboss.org/schema/arquillian" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="http://jboss.org/schema/arquillian http://jboss.org/schema/arquillian/arquillian_1_0.xsd">

    <container qualifier="jboss" default="true" >
        <configuration>
            <property name="jbossHome">${jboss.home}</property>
            <property name="javaVmArguments">-server -Xms64m -Xmx512m -Djboss.socket.binding.port-offset=1000</property>
            <property name="serverConfig">standalone.xml</property>
            <property name="managementAddress">127.0.0.1</property>
            <property name="managementPort">10990</property>
            <property name="waitForPorts">10990</property>
            <property name="waitForPortsTimeoutInSeconds">10</property>
            <property name="allowConnectingToRunningServer">true</property>
        </configuration>
    </container>
</arquillian>
```

PR #127, opened 2020-01-02
MP OpenAPI - Adding misc advanced scenarios tests

fabiobrz:
Last chunk of tests for MicroProfile OpenAPI.

Tests cover the following scenarios (as per TP):

* mixed approach to OpenAPI document model generation (e.g.: static file + programming model + annotations)

* Negative scenario about sending conflicting/unexpedted information for HTTP "Accept" header and `format` query parameter

* JWT related Security annotations, see https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/quarkus-dev/QZgqrblsYEc/sgDcnHYlBAAJ

- N/A Link to the passing job is provided - no Jenkins job can execute the profile yet

### DIFF
--- a/microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
+++ b/microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
@@ -1,0 +1,378 @@
+package org.jboss.eap.qe.microprofile.openapi.advanced;
+
+import static io.restassured.RestAssured.get;
+import static io.restassured.RestAssured.given;
+
+import java.net.URISyntaxException;
+import java.net.URL;
+
+import javax.ws.rs.core.MediaType;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.as.arquillian.api.ServerSetup;
+import org.jboss.as.arquillian.api.ServerSetupTask;
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.eap.qe.microprofile.openapi.OpenApiServerConfiguration;
+import org.jboss.eap.qe.microprofile.openapi.apps.routing.provider.ProviderApplication;
+import org.jboss.eap.qe.microprofile.openapi.apps.routing.provider.RoutingServiceConstants;
+import org.jboss.eap.qe.microprofile.openapi.apps.routing.provider.api.DistrictService;
+import org.jboss.eap.qe.microprofile.openapi.apps.routing.provider.data.DistrictEntity;
+import org.jboss.eap.qe.microprofile.openapi.apps.routing.provider.model.District;
+import org.jboss.eap.qe.microprofile.openapi.apps.routing.provider.rest.DistrictsResource;
+import org.jboss.eap.qe.microprofile.openapi.apps.routing.provider.services.InMemoryDistrictService;
+import org.jboss.eap.qe.microprofile.openapi.v10.OpenApi10OnJaxRsAnnotationsTest;
+import org.jboss.eap.qe.microprofile.tooling.server.configuration.creaper.ManagementClientProvider;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.extras.creaper.core.online.OnlineManagementClient;
+
+/**
+ * Test cases to verify negative scenario in which "Accept" HTTP header and {@code format} query parameter are
+ * conflicting or not valid.
+ * {@link MediaType} constants are used as HTTP "Accept" header values.
+ */
+@RunWith(Arquillian.class)
+@ServerSetup({ AcceptHeadersAndFormatParameterInteractionTest.OpenApiExtensionSetup.class })
+@RunAsClient
+public class AcceptHeadersAndFormatParameterInteractionTest {
+
+    private final static String DEPLOYMENT_NAME = OpenApi10OnJaxRsAnnotationsTest.class.getSimpleName();
+
+    static OnlineManagementClient onlineManagementClient;
+
+    @Deployment(testable = false)
+    public static Archive<?> createDeployment() {
+        return ShrinkWrap.create(
+                WebArchive.class,
+                String.format("%s.war", DEPLOYMENT_NAME))
+                .addClasses(
+                        ProviderApplication.class,
+                        District.class,
+                        DistrictEntity.class,
+                        DistrictService.class,
+                        InMemoryDistrictService.class,
+                        DistrictsResource.class,
+                        RoutingServiceConstants.class)
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+    }
+
+    static class OpenApiExtensionSetup implements ServerSetupTask {
+
+        @Override
+        public void setup(ManagementClient managementClient, String containerId) throws Exception {
+            //  MP OpenAPI up
+            onlineManagementClient = ManagementClientProvider.onlineStandalone();
+            OpenApiServerConfiguration.enableOpenApi(onlineManagementClient);
+        }
+
+        @Override
+        public void tearDown(ManagementClient managementClient, String containerId) throws Exception {
+            //  MP OpenAPI down
+            try {
+                OpenApiServerConfiguration.disableOpenApi(onlineManagementClient);
+            } finally {
+                onlineManagementClient.close();
+            }
+        }
+    }
+
+    /**
+     * @tpTestDetails Sends HTTP request with {@link MediaType#APPLICATION_JSON} {@code Accept} header and no {@code format}
+     *                query parameter.
+     * @tpPassCrit Despite missing {@code format} query parameter, the right Content Type is returned as per
+     *             MicroProfile OpenAPI spec.
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testResponseJsonContentTypeWhenAcceptHeaderAndNoQueryParameter(@ArquillianResource URL baseURL)
+            throws URISyntaxException {
+        given()
+                .header("Accept", MediaType.APPLICATION_JSON)
+                .get(baseURL.toURI().resolve("/openapi"))
+                .then()
+                .statusCode(200)
+                .contentType(MediaType.APPLICATION_JSON);
+    }
+
+    /**
+     * @tpTestDetails Sends HTTP request with {@code format} query parameter.
+     * @tpPassCrit The {@code format} query parameter is used and the right Content Type is returned as per
+     *             MicroProfile OpenAPI spec.
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testResponseJsonContentTypeWhenJsonQueryParameter(@ArquillianResource URL baseURL) throws URISyntaxException {
+        get(baseURL.toURI().resolve("/openapi?format=JSON"))
+                .then()
+                .statusCode(200)
+                .contentType(MediaType.APPLICATION_JSON);
+    }
+
+    /**
+     * @tpTestDetails Sends HTTP request with {@link MediaType#APPLICATION_JSON} {@code Accept} header and {@code YAML} value
+     *                for {@code format} query parameter.
+     * @tpPassCrit Despite conflicting information is sent to server through {@code format} query parameter,
+     *             the right Content Type is returned as per MicroProfile OpenAPI spec.
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testResponseJsonContentTypeWhenAcceptHeaderAndYamlQueryParameter(@ArquillianResource URL baseURL)
+            throws URISyntaxException {
+        given()
+                .header("Accept", MediaType.APPLICATION_JSON)
+                .get(baseURL.toURI().resolve("/openapi?format=YAML"))
+                .then()
+                .statusCode(200)
+                .contentType(MediaType.APPLICATION_JSON);
+    }
+
+    /**
+     * @tpTestDetails Sends HTTP request with {@link MediaType#TEXT_PLAIN} {@code Accept} header and no {@code format} query
+     *                parameter.
+     * @tpPassCrit Despite unexpected information is sent to server through the "Accept" HTTP header,
+     *             the right Content Type is returned as per MicroProfile OpenAPI spec.
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testResponseYamlContentTypeWhenTextPlainAcceptHeaderAndNoQueryParameter(@ArquillianResource URL baseURL)
+            throws URISyntaxException {
+        given()
+                .header("Accept", MediaType.TEXT_PLAIN)
+                .get(baseURL.toURI().resolve("/openapi"))
+                .then()
+                .statusCode(200)
+                .contentType("application/yaml");
+    }
+
+    /**
+     * @tpTestDetails Sends HTTP request with {@link MediaType#TEXT_XML} {@code Accept} header and no {@code format} query
+     *                parameter.
+     * @tpPassCrit Despite unexpected information is sent to server through the "Accept" HTTP header,
+     *             the right Content Type is returned as per MicroProfile OpenAPI spec.
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testResponseYamlContentTypeWhenTextXmlAcceptHeaderAndNoQueryParameter(@ArquillianResource URL baseURL)
+            throws URISyntaxException {
+        given()
+                .header("Accept", MediaType.TEXT_XML)
+                .get(baseURL.toURI().resolve("/openapi"))
+                .then()
+                .statusCode(200)
+                .contentType("application/yaml");
+    }
+
+    /**
+     * @tpTestDetails Sends HTTP request with {@link MediaType#TEXT_HTML} {@code Accept} header and no {@code format} query
+     *                parameter.
+     * @tpPassCrit Despite unexpected information is sent to server through the "Accept" HTTP header,
+     *             the right Content Type is returned as per MicroProfile OpenAPI spec.
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testResponseYamlContentTypeWhenTextHtmlAcceptHeaderAndNoQueryParameter(@ArquillianResource URL baseURL)
+            throws URISyntaxException {
+        given()
+                .header("Accept", MediaType.TEXT_HTML)
+                .get(baseURL.toURI().resolve("/openapi"))
+                .then()
+                .statusCode(200)
+                .contentType("application/yaml");
+    }
+
+    /**
+     * @tpTestDetails Sends HTTP request with {@link MediaType#APPLICATION_XML} {@code Accept} header and no {@code format}
+     *                query parameter.
+     * @tpPassCrit Despite unexpected information is sent to server through the "Accept" HTTP header,
+     *             the right Content Type is returned as per MicroProfile OpenAPI spec.
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testResponseYamlContentTypeWhenApplicationXmlAcceptHeaderAndNoQueryParameter(@ArquillianResource URL baseURL)
+            throws URISyntaxException {
+        given()
+                .header("Accept", MediaType.APPLICATION_XML)
+                .get(baseURL.toURI().resolve("/openapi"))
+                .then()
+                .statusCode(200)
+                .contentType("application/yaml");
+    }
+
+    /**
+     * @tpTestDetails Sends HTTP request with {@link MediaType#APPLICATION_ATOM_XML} {@code Accept} header and no {@code format}
+     *                query parameter.
+     * @tpPassCrit Despite unexpected information is sent to server through the "Accept" HTTP header,
+     *             the right Content Type is returned as per MicroProfile OpenAPI spec.
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testResponseYamlContentTypeWhenAtomXmlAcceptHeaderAndNoQueryParameter(@ArquillianResource URL baseURL)
+            throws URISyntaxException {
+        given()
+                .header("Accept", MediaType.APPLICATION_ATOM_XML)
+                .get(baseURL.toURI().resolve("/openapi"))
+                .then()
+                .statusCode(200)
+                .contentType("application/yaml");
+    }
+
+    /**
+     * @tpTestDetails Sends HTTP request with {@link MediaType#APPLICATION_XHTML_XML} {@code Accept} header and no
+     *                {@code format} query parameter.
+     * @tpPassCrit Despite unexpected information is sent to server through the "Accept" HTTP header,
+     *             the right Content Type is returned as per MicroProfile OpenAPI spec.
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testResponseYamlContentTypeWhenXhtmlXmlAcceptHeaderAndNoQueryParameter(@ArquillianResource URL baseURL)
+            throws URISyntaxException {
+        given()
+                .header("Accept", MediaType.APPLICATION_XHTML_XML)
+                .get(baseURL.toURI().resolve("/openapi"))
+                .then()
+                .statusCode(200)
+                .contentType("application/yaml");
+    }
+
+    /**
+     * @tpTestDetails Sends HTTP request with {@link MediaType#APPLICATION_SVG_XML} {@code Accept} header and no {@code format}
+     *                query parameter.
+     * @tpPassCrit Despite unexpected information is sent to server through the "Accept" HTTP header,
+     *             the right Content Type is returned as per MicroProfile OpenAPI spec.
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testResponseYamlContentTypeWhenSvgXmlAcceptHeaderAndNoQueryParameter(@ArquillianResource URL baseURL)
+            throws URISyntaxException {
+        given()
+                .header("Accept", MediaType.APPLICATION_SVG_XML)
+                .get(baseURL.toURI().resolve("/openapi"))
+                .then()
+                .statusCode(200)
+                .contentType("application/yaml");
+    }
+
+    /**
+     * @tpTestDetails Sends HTTP request with {@link MediaType#SERVER_SENT_EVENTS} {@code Accept} header and no {@code format}
+     *                query parameter.
+     * @tpPassCrit Despite unexpected information is sent to server through the "Accept" HTTP header,
+     *             the right Content Type is returned as per MicroProfile OpenAPI spec.
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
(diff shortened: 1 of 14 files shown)


microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
```diff
@@ -0,0 +1,340 @@
...
+            try {
+                OpenApiServerConfiguration.disableOpenApi(onlineManagementClient);
+            } finally {
+                onlineManagementClient.close();
+            }
+        }
+    }
+
+    /**
+     * @tpTestDetails Sends appropriate "Accept" HTTP header without {@code format} query parameter.
```
istraka: Bad wording. `Sends HTTP request with {@link MediaType.APPLICATION_JSON} header and no {@code format} query parameter` would be better. With no technical background I would read the test sends just a HTTP header without query parameter which makes no sense.
Next thing - what is appropriate?

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
```diff
@@ -0,0 +1,340 @@
...
+    public void testResponseJsonContentTypeWhenAcceptHeaderAndNoQueryParameter(@ArquillianResource URL baseURL)
+            throws URISyntaxException {
+        given().header("Accept", MediaType.APPLICATION_JSON).get(baseURL.toURI().resolve("/openapi"))
+                .then()
+                .statusCode(200)
+                .contentType(MediaType.APPLICATION_JSON);
+    }
+
+    /**
+     * @tpTestDetails Sends {@code format} query parameter.
```
istraka: Wording - `Sends HTTP request with...`.
Similar wording issues are in other JavaDocs too. Would you take a look at those too, please?

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/HybridDocumentationTest.java
```diff
@@ -0,0 +1,240 @@
...
+            configuredHTTPPort = OpenApiServerConfiguration.getHTTPPort(client);
+        }
+
+        //  OpenAPI doc here is exposing a document generated from static file (META-INF/openapi.yaml) plus annotations
+        //  from Local Service Provider "non-routed" JAX-RS endpoints, edited through and OASModelReader implementation
+        //  and eventually filtered through a OASFilter one.
+        //  MP Config is used to tell MP OpenAPI to skip doc generation for exposed "routed" JAX-RS endpoints as service
+        //  consumers must rely to the original documentation resulting from above mentioned original static file modifications.
+        //  Here we also add config properties to reach the Services Provider.
+        String mpConfigProperties = "mp.openapi.scan.exclude.packages=org.jboss.eap.qe.microprofile.openapi.apps.routing.router.rest.routed"
```
istraka: This is not very readable. Can you move the content to the resource file?

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/JWTSecurityAnnotationsTest.java
```diff
@@ -0,0 +1,184 @@
...
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+        return deployment;
+    }
+
+    @Deployment(name = ROUTER_DEPLOYMENT_NAME, order = 2, testable = false)
+    public static Archive<?> localServicesRouterDeployment() {
+        //  OpenAPI doc here is exposing doc generated only from Local Service Provider "non-routed" JAX-RS endpoints
+        //  as MP Config is used to tell MP OpenAPI to skip doc generation for exposed "routed" JAX-RS endpoints
+        //  Here we also add config properties to reach the Services Provider.
+        String mpConfigProperties = "mp.openapi.scan.exclude.packages=org.jboss.eap.qe.microprofile.openapi.apps.routing.router.rest.routed"
```
istraka: Can you move the content to a resource file?

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/JWTSecurityAnnotationsTest.java
```diff
@@ -0,0 +1,184 @@
...
+    @SuppressWarnings("unchecked")
+    public void testForSecurityAnnotationsForAllOperations(
+            @ArquillianResource @OperateOnDeployment(ROUTER_DEPLOYMENT_NAME) URL baseURL) throws URISyntaxException {
+        String responseContent = get(baseURL.toURI().resolve("/openapi"))
+                .then()
+                .statusCode(200)
+                .extract().asString();
+
+        Yaml yaml = new Yaml();
+        Object yamlObject = yaml.load(responseContent);
```
istraka: Can you directly cast the Object to `Map<String, Object>` and create `yamlMap` here?

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
fabiobrz: "appropriate" means "what is expected to be" - i.e. "application/json".

Anyway, I agree about improving the wording, thus fixing...

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
fabiobrz: Fixing.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/HybridDocumentationTest.java
fabiobrz: I understand your point. But there's dynamic configuration involved and I wouldn't like to introduce configuration complexity for a bad worded or misplaced comment. If you agree I'd like to add this comment contents to the deployment method Javadoc. WDYT?

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/JWTSecurityAnnotationsTest.java
fabiobrz: Same as above: my proposal is to add JavaDoc to the deployment method.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/JWTSecurityAnnotationsTest.java
fabiobrz: Sure, you're right. Actually `load` is returning `Map` already, so no need to cast.

fabiobrz:
Hi @istraka, thanks for your review. I addressed JavaDoc related comments but I'd ask your opinion on those two comments about configuration properties. Please let me know whether my proposal makes sense to you too, so that I can push further changes.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/HybridDocumentationTest.java
istraka: Yeah, I have missed those lines at the end. Yeah a method would be better than a resource file. And what about static string and then just use `String.format`? Some memory and processor time will be saved. WDYT?

fabiobrz: yes, I do agree. Fixing...

fabiobrz:
Hi @istraka, thanks. Addressed your last comments.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
```diff
@@ -0,0 +1,365 @@
...
+import org.wildfly.extras.creaper.core.online.OnlineManagementClient;
+
+/**
+ * Test cases to verify negative scenario in which "Accept" HTTP header and {@code format} query parameter are
+ * conflicting or not valid.
+ * {@link MediaType} constants are used as HTTP "Accept" header values.
+ */
+@RunWith(Arquillian.class)
+@ServerSetup({ AcceptHeadersAndFormatParameterInteractionTest.OpenApiExtensionSetup.class })
+@RunAsClient
```
mnovak1: As all tests are `@RunAsClient` and test itself is not deployed into server with deployment there is no need to use `@ServerSetup`. Could you modify the test so setup() and tearDown() are annotated by standard JUnit `@BeforeClass` and `@AfterClass` annotation (and moved from inner class), please?

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
```diff
@@ -0,0 +1,365 @@
...
+            }
+        }
+    }
+
+    /**
+     * @tpTestDetails Sends HTTP request with {@link MediaType#APPLICATION_JSON} {@code Accept} header and no {@code format}
+     *                query parameter.
+     * @tpPassCrit Despite missing {@code format} query parameter, the right Content Type is returned as per
+     *             MicroProfile OpenAPI spec.
+     *
```
mnovak1: Could you remove this line in all JavaDocs, please?

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
```diff
@@ -0,0 +1,365 @@
...
+     * @tpSince EAP 7.4.0.CD19
+     */
+    @Test
+    public void testResponseJsonContentTypeWhenAcceptHeaderAndNoQueryParameter(@ArquillianResource URL baseURL)
+            throws URISyntaxException {
+        given().header("Accept", MediaType.APPLICATION_JSON).get(baseURL.toURI().resolve("/openapi"))
+                .then()
+                .statusCode(200)
+                .contentType(MediaType.APPLICATION_JSON);
```
mnovak1: Could you update code formatting as it's in [1], please? It will unite formatting for MP OpenAPI tests and looks nicer to me :-)

[1] https://github.com/jboss-eap-qe/eap-microprofile-test-suite/blob/master/microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/v11/MicroProfileOpenApi11Test.java#L99

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/HybridDocumentationTest.java
```diff
@@ -0,0 +1,252 @@
...
+ * Test to assess that the process of generating OpenAPI document is properly executed by following all the
+ * MicroProfile OpenAPI spec processing rules - i.e. starting from static file and performing customization and
+ * filtering.
+ * The real world scenario for Local Services Router app is implemented here: the OpenAPI document is provided as
+ * deliverable by Service Provider staff to Local Services Router staff so that it can be used as official documentation
+ * once proper customization is applied.
+ */
+@RunWith(Arquillian.class)
+@ServerSetup({ HybridDocumentationTest.OpenApiExtensionSetup.class })
+@RunAsClient
```
mnovak1: Looks like the same thing as above, `ServerSetup` is not needed.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/JWTSecurityAnnotationsTest.java
```diff
@@ -0,0 +1,211 @@
...
+import org.junit.runner.RunWith;
+import org.wildfly.extras.creaper.core.online.OnlineManagementClient;
+import org.yaml.snakeyaml.Yaml;
+
+/**
+ * Test cases for usage of JWT related security MP OpenAPI annotations
+ */
+@RunWith(Arquillian.class)
+@ServerSetup({ JWTSecurityAnnotationsTest.OpenApiExtensionSetup.class })
+@RunAsClient
```
mnovak1: Looks like the same thing as above, `ServerSetup` is not needed.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
```diff
@@ -0,0 +1,365 @@
...
+                .addClasses(ProviderApplication.class)
+                .addClasses(
+                        District.class,
+                        DistrictEntity.class,
+                        DistrictService.class,
+                        InMemoryDistrictService.class,
+                        DistrictsResource.class,
+                        RoutingServiceConstants.class)
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+        return deployment;
```
mnovak1: Optional: there is no need to create variable `deployment` and `WebArchive`object can be retruned directly. I can see it in other test cases as well.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/JWTSecurityAnnotationsTest.java
mnovak1: @fabiobrz I can see the same thing in other test classes for example `MicroProfileOpenApi11Test`, do you think you might change it there as well, please?

mnovak1:
@fabiobrz I did review of the PR and have mostly formatting issues + use of `@ServerSetup` task which does not seem to be necessary.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
fabiobrz: Well, actually that usage of `ServerSetupTask` is IMO _very_ necessary, at least for MP OpenAPI tests, because it is used to properly control the _server_ configuration with respect to the managed deployment life cycle, so it is not a discriminant the fact that the test is not deployed at server side, see JavaDoc at [1].
I discussed this on the working group chat a couple of weeks ago, here's an excerpt for you to understand the details of what I ran into:

```
...
My issue:
* I was having two warnings from Weld because it could not link two registered MP OpenAPI extension classes implementing OASModelReader and OASFilter interfaces for programmatic OpenAPI doc manipulation 
* first WARN pops up when Arquillian archive is deployed => possible root cause: @BeforeClass happens later than deploy and I enable openapi extension/subsystem in @BeforeClass, so: a deployment contains and registers concrete Filter/ModelReder, Weld tries and load them but the extension is still disabled. If such reasoning is true, enabling the extension should happen soon before deployment.
* second (and last) warn pops up when calling CLI reload command. This happens in @AfterClass, soon after disabling the extension. This means that Weld processes again the - still undeployed deployment containing registered implementations for Filter and ModelReader and emits the same warning. This means that disabling the extension should happen soon after undeployment.
...
```

Similar Arquillian deployment life cycle management issues were observed in some different test cases (for one of those I filed a WIP issue which was solved with this configuration improvement) and were all related to subsystem configuration code being executed in `@BeforeClass` and `@AfterClass`.

Bottom line, enabling and/or disabling the subsystem is - at least for MP OpenAPI spec - a server configuration task that must be synchronous with related server extension life cycle management activities that could affect the deployment phase (e.g.: usage of registered interface implementation as in the above mentioned use case).

I have discussed the thing with developer, which helped in analyzing what was going on and I made some research in Wildfly TS too, where `ServerSetupTask` is used to configure server even when tests are to be run by the client. 

Finally, once that all of the above mentioned rationale was validated, I applied the configuration to all of the MP OpenAPI tests in order to follow the same pattern throughout the module. 

So, to be honest I don't see any value for changing this logical pattern with JUint `@BeforeClass` and `@AfterCless` methods and indeed I believe this is actually unwanted.

Let me know your thoughts.

[1]
https://github.com/wildfly/wildfly-arquillian/blob/master/common/src/main/java/org/jboss/as/arquillian/api/ServerSetupTask.java

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
fabiobrz: Ok, will do.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
fabiobrz: Sure, will do.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/HybridDocumentationTest.java
fabiobrz: See my comment at https://github.com/jboss-eap-qe/eap-microprofile-test-suite/pull/127#discussion_r362746831

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/JWTSecurityAnnotationsTest.java
fabiobrz: See my comment at https://github.com/jboss-eap-qe/eap-microprofile-test-suite/pull/127#discussion_r362746831

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
fabiobrz: Yes, thanks, fixing.

fabiobrz:
Hi @mnovak1 thanks for your review. I'll be fixing the formatting issues right now. Please read my explanation about ServerSetupTask and let me know your thoughts.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
mnovak1: @fabiobrz Thanks for explanation, now I can see the issue and the motivation for it. So when deployment is managed by Arquillian then it's deployed before `@BeforeClass` is called (so before OpenApi is enabled). 

However the issue can be workaround by manual deployment like:
```
@ArquillianResource
private Deployer deployer;

@Deployment(name = FIRST_DEPLOYMENT, managed = false)
...

 @BeforeClass
public static void setup() throws ManagementClientRelatedException {
        onlineManagementClient = ManagementClientProvider.onlineStandalone();
        OpenApiServerConfiguration.enableOpenApi(onlineManagementClient);
        deployer.deploy(FIRST_DEPLOYMENT);
}
```

I'm unhappy that `managementClient` is passed as parameter into setup(...) method and then it's not used and instead there is used Creaper `onlineManagementClient`:
```
static class OpenApiExtensionSetup implements ServerSetupTask {

        @Override
        public void setup(ManagementClient managementClient, String containerId) throws Exception {
            //  MP OpenAPI up
            onlineManagementClient = ManagementClientProvider.onlineStandalone();
            OpenApiServerConfiguration.enableOpenApi(onlineManagementClient);
        }
```
the code looks weird. 

@rsvoboda @istraka wdyt?

fabiobrz: Hi @mnovak1 - thanks for your reply, see my comments here.

> However the issue can be workaround by manual deployment like:
...

Well yes, it possibly could. But I am thinking why it should, as `ServerSetupTask` implementation is the tool that Arquillian is providing to the client in order to manage server configuration task _within_ the logical scope of it's (Arquillian's) managed deployment process. So I am trying and find any reason for changing it all again in favor of a more generic - i.e. JUnit - implementation. But feel free to let me know any rationale for it.   

> I'm unhappy that managementClient is passed as parameter into setup(...) method and then it's not used and instead there is used Creaper onlineManagementClient:

So you're suggesting not to use our tooling for server configuration here because a `ManagementClient` instance is already injected? I am actually not sure that the API will prove to be sufficient for our current requirements - as our tooling has proved to be, but I have actually not tried. WDYT?

fabiobrz:
Hi @mnovak1, I've pushed changes for formatting issues and resolved all related conversations, while I left the one about server setup open so that we can discuss it further.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/security/ListenerSecurityConfigurationTest.java
```diff
@@ -84,30 +85,27 @@ public static WebArchive createServicesProviderDeployment() {
...
-        return deployment;
     }
 
     static class OpenApiExtensionSetup implements ServerSetupTask {
 
         @Override
         public void setup(ManagementClient managementClient, String containerId) throws Exception {
             jbossHome = arquillianContainerProperties.getContainerProperty("jboss", "jbossHome");
-            onlineManagementClient = ManagementClientProvider.onlineStandalone();
+            listenersConfigurationOnlineManagementClient = ManagementClientProvider.onlineStandalone();
```
mnovak1: Could be `listenersConfigurationOnlineManagementClient` initialized in `@BeforeClass`, please?

fabiobrz: Yes, sure, sorry about this, I definitely missed it.

fabiobrz: Done

fabiobrz:
Hi @mnovak1, thanks for your work. I've fixed your last comment.

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/AcceptHeadersAndFormatParameterInteractionTest.java
fabiobrz: After discussion in chat, current implementation is now using the following wrapping logic provided by new method in `ManagementClientProvider`:

```
...
org.wildfly.extras.creaper.core.ManagementClient.onlineLazy(
                OnlineOptions.standalone().wrap(managementClient.getControllerClient()))
...
```

So `ServerSetupTask#setup()` and `ServerSetupTask#tearDown()` methods implementation can use the `ManagementClient` instance passed as input parameter in order to get a Creaper `OnlineManagmentClient` out of Arquillian `ManagementClient` instance.
This is more direct and clean, although eventually relying on the same configuration source - i.e. `arquillian.xml`.

Due to the following behavior, explained by the `OnlineOptions.standalone().wrap()` Javadoc:
```
         * <p>The {@link org.wildfly.extras.creaper.core.online.OnlineManagementClient} that will be created will wrap
         * the provided {@link org.jboss.as.controller.client.ModelControllerClient modelControllerClient}.</p>
         *
         * <p><b>Note that doing this ties together the lifecycles of the {@code OnlineManagementClient} and
         * the wrapped {@code ModelControllerClient}!</b> If you close the {@code ModelControllerClient}, the
         * {@code OnlineManagementClient} becomes unusable. If you close the {@code OnlineManagementClient},
         * the wrapped {@code ModelControllerClient} will be closed as well.</p>
         *
         * <p>This also means that the new {@code OnlineManagementClient} <b>can't be {@code reconnect}ed</b>, which
         * is sometimes needed if server reload or restart is performed. Hence, use client wrapping with caution!</p>
```
the `OnlineManagmentClient` instance obtained as a wrapper for `ModelControllerClient` instance is never closed, either implicitly or explicitly: e.g.:

```
static class OpenApiExtensionSetup implements ServerSetupTask {

        @Override
        public void setup(ManagementClient managementClient, String containerId) throws Exception {
            OpenApiServerConfiguration.enableOpenApi(ManagementClientProvider.onlineStandalone(managementClient));
        }

        @Override
        public void tearDown(ManagementClient managementClient, String containerId) throws Exception {
            OpenApiServerConfiguration.disableOpenApi(ManagementClientProvider.onlineStandalone(managementClient));
        }
    }
```
as this would cause the underlying `ModelControllerClient` instance too, which would affect Arquillian configuration globally.

mnovak1: @fabiobrz thanks Fabio, I don't think closing is left on GC but on `@ServerSetupTask` but this would need closer look. The main point here that it must be closed manually in the test which is common thing when using `@ServerSetupTask`

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/HybridDocumentationTest.java
```diff
@@ -0,0 +1,239 @@
...
+                .addClasses(
+                        RouterApplication.class,
+                        LocalServiceRouterInfoResource.class)
+                .addClasses(
+                        DistrictObject.class,
+                        RouterDistrictsResource.class,
+                        DistrictServiceClient.class)
+                .addClasses(
+                        OpenApiModelReader.class,
+                        OpenApiFilter.class)
```
mnovak1: Could you join the `addClasses` here, please?

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/advanced/JWTSecurityAnnotationsTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+        return ShrinkWrap.create(
+                WebArchive.class, ROUTER_DEPLOYMENT_NAME + ".war")
+                .addClasses(
+                        ProviderApplication.class,
+                        LocalServiceRouterInfoResource.class)
+                .addClasses(
+                        DistrictObject.class,
+                        RouterDistrictsResource.class,
+                        DistrictServiceClient.class)
```
mnovak1: Same here (join addClasses())

microprofile-open-api/src/test/java/org/jboss/eap/qe/microprofile/openapi/integration/config/MicroprofileConfigIntegrationTests.java
```diff
@@ -73,11 +80,17 @@ public static WebArchive serviceProviderDeployment() {
                         DistrictsResource.class,
                         RoutingServiceConstants.class)
```
mnovak1: And here.

mnovak1:
@fabiobrz I've added one last thing to correct, othwere good to merge for me. Also could you rebase, please?

fabiobrz:
Hi @mnovak1 thanks, I've pushed the changes you suggested.

mnovak1:
@fabiobrz thanks, merging

PR #156, opened 2020-01-09
MP Health test development

istraka:
https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/eap-7.x-microprofile-testsuite/45/

Fixes https://github.com/jboss-eap-qe/eap-microprofile-test-suite/issues/162

### DIFF
--- a/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
+++ b/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
@@ -6,7 +6,6 @@ import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.hasItems;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.not;
 
 import java.io.IOException;
 import java.util.Collections;
@@ -16,6 +15,8 @@ import java.util.concurrent.TimeoutException;
 
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.as.arquillian.api.ServerSetup;
 import org.jboss.eap.qe.microprofile.health.tools.HealthUrlProvider;
 import org.jboss.eap.qe.microprofile.tooling.server.configuration.ConfigurationException;
 import org.jboss.eap.qe.microprofile.tooling.server.configuration.arquillian.ArquillianContainerProperties;
@@ -23,11 +24,13 @@ import org.jboss.eap.qe.microprofile.tooling.server.configuration.arquillian.Arq
 import org.jboss.eap.qe.microprofile.tooling.server.configuration.creaper.ManagementClientProvider;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Ignore;
 import org.junit.Test;
+import org.junit.runner.RunWith;
 import org.wildfly.extras.creaper.core.online.OnlineManagementClient;
 import org.wildfly.extras.creaper.core.online.operations.admin.Administration;
 
@@ -35,6 +38,9 @@ import io.restassured.http.ContentType;
 import io.restassured.specification.RequestSpecification;
 
 /**
+ * MP FT integration will be tested relying on microprofile-config.properties,
+ * MP Config properties are defined in microprofile-config.properties in META INF.
+ *
  * Testclass to test {@code /health}, {@code /health/live}, {@code /health/ready} endpoints.
  * {@link CDIBasedLivenessHealthCheck} and {@link CDIBasedReadinessHealthCheck} probes are based on fail-safe CDI bean
  * {@link FailSafeDummyService} configured at the beginning via MP Config properties:
@@ -47,13 +53,18 @@ import io.restassured.specification.RequestSpecification;
  * </ul>
  * If service is in maintenance a method simulating opening resources {@link FailSafeDummyService#simulateOpeningResources}
  * throws a {@link IOException}.
- * Subclasses are supposed to configure only MP Config - set MP Config properties in different ways.
  *
- * Tests in this test class are not expected to change the state of probes after their initialization since some
- * ways of configuring MP Config would require a Wildfly reload in order to update those values.
  */
+@RunWith(Arquillian.class)
 @RunAsClient
-public abstract class FailSafeCDIHealthBaseTest {
+@ServerSetup(MicroProfileFTSetupTask.class)
+public class FailSafeCDIConfigFileHealthTest {
+
+    private static final String MPConfigContent = String.format("%s=%s\n%s=%s\n%s=%s\n%s=%s",
+            FailSafeDummyService.LIVE_CONFIG_PROPERTY, false,
+            FailSafeDummyService.READY_CONFIG_PROPERTY, true,
+            FailSafeDummyService.IN_MAINTENANCE_CONFIG_PROPERTY, true,
+            FailSafeDummyService.READY_IN_MAINTENANCE_CONFIG_PROPERTY, false);
 
     protected static RequestSpecification metricsRequest;
 
@@ -79,96 +90,8 @@ public abstract class FailSafeCDIHealthBaseTest {
         return ShrinkWrap
                 .create(WebArchive.class, FailSafeCDIHealthBaseTest.class.getSimpleName() + ".war")
                 .addClasses(FailSafeDummyService.class, CDIBasedLivenessHealthCheck.class, CDIBasedReadinessHealthCheck.class)
+                .addAsManifestResource(new StringAsset(MPConfigContent), "microprofile-config.properties")
                 .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
-    }
-
-    /**
-     * Subclass is expected to implement the method to set the MP Config properties in a way specific to MP Config
-     * configuration.
-     */
-    protected abstract void setConfigProperties(boolean live, boolean ready, boolean inMaintenance, boolean readyInMaintenance)
-            throws Exception;
-
-    /**
-     * @tpTestDetails Multiple-component customer scenario - health check status is based on fail-safe CDI bean and MP Config
-     *                property.
-     * @tpPassCrit Overall and the health check status are up. MP Metrics are increased according to the specification.
-     * @tpSince EAP 7.4.0.CD19
-     */
-    @Test
-    final public void testHealthEndpointUp() throws Exception {
-        setConfigProperties(true, true, false, false);
-        get(HealthUrlProvider.healthEndpoint()).then()
-                .contentType(ContentType.JSON)
-                .statusCode(200)
-                .body("status", is("UP"),
-                        "checks", hasSize(2),
-                        "checks.status", hasItems("UP"),
-                        "checks.status", not(hasItems("DOWN")),
-                        "checks.name", containsInAnyOrder("dummyLiveness", "dummyReadiness"));
-
-        MetricsChecker.get(metricsRequest)
-                .validateSimulationCounter(1)
-                .validateInvocationsTotal(1)
-                .validateRetryCallsSucceededNotTriedTotal(1);
-
-        // same request have been validated above, now we need to increase metrics
-        get(HealthUrlProvider.healthEndpoint()).then().statusCode(200);
-
-        MetricsChecker.get(metricsRequest)
-                .validateSimulationCounter(2)
-                .validateInvocationsTotal(2)
-                .validateRetryCallsSucceededNotTriedTotal(2);
-    }
-
-    /**
-     * @tpTestDetails Multiple-component customer scenario - health check status is based on fail-safe CDI bean and MP Config
-     *                property.
-     * @tpPassCrit Overall and the health check status are up.
-     * @tpSince EAP 7.4.0.CD19
-     */
-    @Test
-    final public void testLivenessEndpointUp() throws Exception {
-        setConfigProperties(true, true, false, false);
-        get(HealthUrlProvider.liveEndpoint()).then()
-                .statusCode(200)
-                .contentType(ContentType.JSON)
-                .body("status", is("UP"),
-                        "checks", hasSize(1),
-                        "checks.status", hasItems("UP"),
-                        "checks.status", not(hasItems("DOWN")),
-                        "checks.name", containsInAnyOrder("dummyLiveness"));
-    }
-
-    /**
-     * @tpTestDetails Multiple-component customer scenario - health check status is based on fail-safe CDI bean and MP Config
-     *                property.
-     * @tpPassCrit Overall and the health check status are up. MP Metrics are increased according to the specification.
-     * @tpSince EAP 7.4.0.CD19
-     */
-    @Test
-    final public void testReadinessEndpointUp() throws Exception {
-        setConfigProperties(true, true, false, false);
-        get(HealthUrlProvider.readyEndpoint()).then()
-                .statusCode(200)
-                .contentType(ContentType.JSON)
-                .body("status", is("UP"),
-                        "checks", hasSize(1),
-                        "checks.status", hasItems("UP"),
-                        "checks.name", containsInAnyOrder("dummyReadiness"));
-
-        MetricsChecker.get(metricsRequest)
-                .validateSimulationCounter(1)
-                .validateInvocationsTotal(1)
-                .validateRetryCallsSucceededNotTriedTotal(1);
-
-        // same request have been validated above, now we need to increase metrics
-        get(HealthUrlProvider.readyEndpoint()).then().statusCode(200);
-
-        MetricsChecker.get(metricsRequest)
-                .validateSimulationCounter(2)
-                .validateInvocationsTotal(2)
-                .validateRetryCallsSucceededNotTriedTotal(2);
     }
 
     /**
@@ -181,7 +104,6 @@ public abstract class FailSafeCDIHealthBaseTest {
     @Ignore("WFLY-12924, WFLY-12925")
     @Test
     final public void testHealthEndpointDownInMaintenance() throws Exception {
-        setConfigProperties(false, true, true, false);
         // TODO Java 11 Map<String, String> liveCheck = Map.of( "name", "dummyLiveness", "status", "DOWN");
         Map<String, String> liveCheck = Collections.unmodifiableMap(new HashMap<String, String>() {
             {
@@ -232,7 +154,6 @@ public abstract class FailSafeCDIHealthBaseTest {
      */
     @Test
     final public void testLivenessEndpointDownInMaintenance() throws Exception {
-        setConfigProperties(false, true, true, false);
         get(HealthUrlProvider.liveEndpoint()).then()
                 .statusCode(503)
                 .contentType(ContentType.JSON)
@@ -252,7 +173,6 @@ public abstract class FailSafeCDIHealthBaseTest {
     @Ignore("WFLY-12924, WFLY-12925")
     @Test
     final public void testReadinessEndpointDownInMaintenance() throws Exception {
-        setConfigProperties(false, true, true, false);
         get(HealthUrlProvider.readyEndpoint()).then()
                 .statusCode(503)
                 .contentType(ContentType.JSON)
@@ -282,5 +202,4 @@ public abstract class FailSafeCDIHealthBaseTest {
                 .validateRetryCallsSucceededTotal(0)
                 .validateFallbackCallsTotal(2);
     }
-
 }
--- a/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIHealthDynamicBaseTest.java
+++ b/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIHealthDynamicBaseTest.java
@@ -14,6 +14,7 @@ import java.util.Map;
 
 import org.jboss.arquillian.container.test.api.RunAsClient;
 import org.jboss.eap.qe.microprofile.health.tools.HealthUrlProvider;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import io.restassured.http.ContentType;
@@ -21,7 +22,8 @@ import io.restassured.http.ContentType;
 /**
  * Testclass to test {@code /health}, {@code /health/live}, {@code /health/ready} endpoints. Liveness and readiness probes
  * are configured via MP Config properties at the beginning and changed later in test execution. Tests in the testclass
- * expect no reload operation after changing the values.
+ * expect no reload operation after changing the values (during the test execution).
+ * Test class extends {@link FailSafeCDIHealthBaseTest} for those scenarios that expect dynamic MP Config.
  */
 @RunAsClient
 public abstract class FailSafeCDIHealthDynamicBaseTest extends FailSafeCDIHealthBaseTest {
@@ -33,7 +35,7 @@ public abstract class FailSafeCDIHealthDynamicBaseTest extends FailSafeCDIHealth
      * @tpSince EAP 7.4.0.CD19
      */
     @Test
-    public void testHealthEndpointUpToDown() throws Exception {
+    final public void testHealthEndpointUpToDown() throws Exception {
         setConfigProperties(true, true, false, false);
         get(HealthUrlProvider.healthEndpoint()).then()
                 .statusCode(200)
@@ -44,6 +46,11 @@ public abstract class FailSafeCDIHealthDynamicBaseTest extends FailSafeCDIHealth
                         "checks.status", hasItems("UP"),
                         "checks.status", not(hasItems("DOWN")),
                         "checks.name", containsInAnyOrder("dummyLiveness", "dummyReadiness"));
+
+        MetricsChecker.get(metricsRequest)
+                .validateSimulationCounter(1)
+                .validateInvocationsTotal(1)
+                .validateRetryCallsSucceededNotTriedTotal(1);
 
         setConfigProperties(true, false, false, true);
         // TODO Java 11 Map<String, String> liveCheck = Map.of( "name", "dummyLiveness", "status", "UP");
@@ -66,6 +73,11 @@ public abstract class FailSafeCDIHealthDynamicBaseTest extends FailSafeCDIHealth
                 .body("status", is("DOWN"),
                         "checks", hasSize(2),
                         "checks", containsInAnyOrder(liveCheck, readyCheck));
+
+        MetricsChecker.get(metricsRequest)
+                .validateSimulationCounter(2)
+                .validateInvocationsTotal(2)
+                .validateRetryCallsSucceededNotTriedTotal(2);
     }
 
     /**
@@ -75,12 +87,11 @@ public abstract class FailSafeCDIHealthDynamicBaseTest extends FailSafeCDIHealth
      * @tpSince EAP 7.4.0.CD19
      */
     @Test
-    public void testLivenessEndpointUpToDown() throws Exception {
+    final public void testLivenessEndpointUpToDown() throws Exception {
         setConfigProperties(true, true, false, false);
         get(HealthUrlProvider.liveEndpoint()).then()
                 .statusCode(200)
                 .contentType(ContentType.JSON)
-                .header("Content-Type", containsString("application/json"))
                 .body("status", is("UP"),
                         "checks", hasSize(1),
                         "checks.status", hasItems("UP"),
@@ -90,7 +101,6 @@ public abstract class FailSafeCDIHealthDynamicBaseTest extends FailSafeCDIHealth
         get(HealthUrlProvider.liveEndpoint()).then()
                 .statusCode(503)
                 .contentType(ContentType.JSON)
-                .header("Content-Type", containsString("application/json"))
                 .body("status", is("DOWN"),
                         "checks", hasSize(1),
                         "checks.status", hasItems("DOWN"),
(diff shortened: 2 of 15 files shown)


microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIModelFilePropsHealthTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.runner.RunWith;
+import org.wildfly.extras.creaper.core.online.OnlineManagementClient;
+import org.wildfly.extras.creaper.core.online.operations.admin.Administration;
+
+/**
+ * MP Config property is provided by file-props model option - values are stored in files on FS.
```
fabiobrz: Here too it would be nice to have some more detail about the logic - e.g.: the test calss would backup the file-props config _before each test_ (is that wanted?) and restore it soon after it has completed,, emphasizing what happens in the middle: MP FT integration will be tested relying on FS based MP Config. Just my 2 cents.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIModelFilePropsHealthTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Override
+    protected void setConfigProperties(boolean live, boolean ready, boolean inMaintanance, boolean readyInMainenance)
+            throws Exception {
+        Files.write(liveFile, Boolean.toString(live).getBytes(StandardCharsets.UTF_8));
+        Files.write(readyFile, Boolean.toString(ready).getBytes(StandardCharsets.UTF_8));
+        Files.write(inMainenanceFile, Boolean.toString(inMaintanance).getBytes(StandardCharsets.UTF_8));
+        Files.write(readyInMainenanceFile, Boolean.toString(readyInMainenance).getBytes(StandardCharsets.UTF_8));
+        new Administration(ManagementClientProvider.onlineStandalone(managementClient)).reload();
+
```
fabiobrz: Additional empty line here?
Anyway, here again I had to go to the superclass implementation to see that `setConfigProperties()` is called before each test and so `new Administration(ManagementClientProvider.onlineStandalone(managementClient)).reload();` is meaningful.
This is IMO another signal for a refactoring in order to get the logical flow implemented differently in the classes/hierarchy.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIModelPropsHealthTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.runner.RunWith;
+import org.wildfly.extras.creaper.core.online.OnlineManagementClient;
+import org.wildfly.extras.creaper.core.online.operations.admin.Administration;
+
+/**
+ * MP Config property is provided by props model option - values are stored in model object.
```
fabiobrz: Final reminder for improving Javadoc for this set of classes, if you agree.

microprofile-health/src/test/resources/org/jboss/eap/qe/microprofile/health/integration/file-props/dummy.in_maintenance
```diff
@@ -0,0 +1 @@
+2
```
fabiobrz: Would please consider the idea of adding a one line comment to make clear that numbers here are nothing else that placeholders or whatever?

istraka: There can not be any coment. The file must contain value for the property which name is equal to the filename

fabiobrz: Sorry but I don't get this. Aren't you providing those values as Boolean?

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.runner.RunWith;
+
+/**
+ * MP Config properties are defined in microprofile-config.properties in META INF. Since properties are immutable,
+ * only some of test cases are valid.
+ */
+@RunWith(Arquillian.class)
+@RunAsClient
+@ServerSetup(MicroProfileFTSetupTask.class)
+public class FailSafeCDIConfigFileHealthTest extends FailSafeCDIHealthBaseTest {
```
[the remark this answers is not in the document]
istraka: The purpose of this design is to have Set of valid testcases - `FailSafeCDIHealthBaseTest` and `FailSafeCDIHealthDynamicBaseTest`. Inheriting classes are supposed to deploy same Liveness and Readiness probes.
I see your point. WDYT if`FailSafeCDIHealthBaseTest` creates deployment and define abstract method for changing it - namely adding `ConfigSource` and `ConfigSourceProvider`?

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jboss.as.arquillian.api.ServerSetup;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.runner.RunWith;
+
+/**
+ * MP Config properties are defined in microprofile-config.properties in META INF. Since properties are immutable,
+ * only some of test cases are valid.
```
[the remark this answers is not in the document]
istraka: Fixing.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public void testLivenessEndpointUp() {
+    }
+
+    @Override
+    public void testReadinessEndpointUp() {
+    }
+
+    @Override
+    protected void setConfigProperties(boolean live, boolean ready, boolean inMaintanance, boolean readyInMainenance) {
```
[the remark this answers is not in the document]
istraka: Fixing

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
```diff
@@ -0,0 +1,58 @@
...
+        if (!(!live && ready && inMaintanance && !readyInMainenance)) {
```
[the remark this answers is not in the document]
istraka: only valid scenario is if service is not live and not ready and in maintenance (ready is true, but that would be valid if the service is not in maintenance). Thing is, MP Config is set this way and can not be changed. Hence other scenarios are invalid.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIModelFilePropsHealthTest.java
istraka: Thanks

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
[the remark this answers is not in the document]
fabiobrz: Thanks @istraka, I can get what your goal is, now. 
That said, if the deployment building business logic is common to the hierarchy then definitely it sounds reasonable to have it encapsulated and provided by the superclass.
Regarding the refactoring, what you suggest seems about right at first sight, but in this case, subclasses would have to be just _providers_ for MP configuration and test methods should have to be defined as `final`, IMO. But that is just one of the viable approaches.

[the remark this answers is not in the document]
istraka: Thanks. I agree but not with test being `final`. There is a way of configuration - microprofile-config.properties in META-INF where changes to properties are not allowed. Thus some scenarios must be Overriden and do nothing.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIModelPropsHealthTest.java
istraka: Do you have any suggestions? Properties are set via cli command and are stored directly in WildFly via model.

microprofile-health/src/test/resources/org/jboss/eap/qe/microprofile/health/integration/file-props/dummy.in_maintenance
istraka: what about `a random string to be changed`?

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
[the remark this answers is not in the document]
fabiobrz: > scenario is if service is not live

Is what you state actually matching `!(!live) == live` condition?

IIUC: if it is _false_ that service is (not live && inMaintenance && ready && not in readyMaintenance) then BAD SCENARIO.

But, anyway, I get your point about the condition that is allowing for this scenario to be considered as "valid", this comment of mine was actually about how easily that conditional statement could be read.

If I think of that scenario I would imagine:
* a live service - which is line && not in maintenance 
AND
* a readiness service - which is _not_ ready && in maintenance

that would translate to something like:
```
Boolean unvalidScenario = !((live && !ready) && (!ready && !readyInMainenance))
if (unvalidScenario) { throw ... }
```
But yeah, I might be missing some business logic and that's just my taste.

microprofile-health/src/test/resources/org/jboss/eap/qe/microprofile/health/integration/file-props/dummy.in_maintenance
fabiobrz: Sounds good to me, anything that would not lead to think about enumeration or some hidden logic.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIModelPropsHealthTest.java
fabiobrz: Well, the Javadoc is correct IMO, I was just pointing out a more general issue with documentation with respect to hierarchy, as per my previous comments regarding the other test classes. 
Just a `reminder`, as I stated.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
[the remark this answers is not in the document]
fabiobrz: Right. But then the issue is still there: if I am allowed to override, I'll do it. And if I do it, then I need to know what to expect to be deployed. But what is deployed is encapsulated by superclass, so here's the mismatch. 
IMO the way you described it in your previous comment, defines the subclasses just as configuration providers, not test case implementors, doesn't it?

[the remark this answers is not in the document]
istraka: Yest. However tests expect that configuration can be set. What about moving all but one tests from `FailSafeCDIHealthBaseTest` to a new class `FailSafeCDIHealthSetttableTest`. `FailSafeCDIHealthBaseTest` class would define one test with expected config properties. `FailSafeCDIConfigFileHealthTest` would inherit from it. We would get rid of the mismatch. A disadvantage is that there will be 3 abstract classes...

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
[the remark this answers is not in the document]
istraka: Maintenance applies only on a readiness probe, that is the catch. However I like what you are proposing, I am gonna change it that way.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIModelFilePropsHealthTest.java
istraka: well that is because different way of configuring MP Config requires different way of setting the properties. Some ways needs reload to apply the changes some does not.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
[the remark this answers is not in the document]
fabiobrz: So, IIUC:
`FailSafeCDIHealthBaseTest` -> one test, with expected config properties. The subclass (`FailSafeCDIConfigFileHealthTest`) for this one are expected to override test methods? In such case how do they know about the deploymnet being built by the superclass?
`FailSafeCDIHealthSettableTest` -> This one would be used by those subclasses not overriding the tests methods, right?
If so, I think you'd have the mismatch for FailSafeCDIHealthBaseTest hierarchy but maybe I didn't get your explanation well.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIModelFilePropsHealthTest.java
fabiobrz: Yes, sure. I get what you mean, a different scenario requires for different actions to be taken. 
My comment here was related to the same argument we mentioned earlier, i.e. inheritance and so on.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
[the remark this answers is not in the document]
istraka: All test would be final
- `FailSafeCDIHealthBaseTest` one test with expected properties (constants). Subclasses are expected to configure MP Config in that only way.
- `FailSafeCDIHealthSettableTest` rest of origin tests (from `FailSafeCDIHealthBaseTest`). Subclasses are expected to be able to set properties after deployment.

Hierarchy change:
- `FailSafeCDIConfigFileHealthTest` inherits from `FailSafeCDIHealthBaseTest`
- other classes would inherit from `FailSafeCDIHealthSettableTest` (those that have extended `FailSafeCDIHealthBaseTest` originally)

istraka:
@fabiobrz  I have addresed issues, please review. At the end I have decided to retract `FailSafeCDIConfigFileHealthTest` from the hierarchy. IMO this is best trade-off.
- not super complicated with 3 abstract classes
- test method are final
- deployment method in top leve abstract class (it would not be possible for `FailSafeCDIConfigFileHealthTest` to easily change it without giving the class "too much power")

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
```diff
@@ -0,0 +1,204 @@
...
+                .create(WebArchive.class, FailSafeCDIHealthBaseTest.class.getSimpleName() + ".war")
+                .addClasses(FailSafeDummyService.class, CDIBasedLivenessHealthCheck.class, CDIBasedReadinessHealthCheck.class)
+                .addAsManifestResource(new StringAsset(MPConfigContent), "microprofile-config.properties")
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+    }
+
+    /**
+     * @tpTestDetails Multiple-component customer scenario - health check status is based on fail-safe CDI bean and MP Config
+     *                property. There is IOException since service is in maintenance but fallback method should be used.
+     * @tpPassCrit Overall and the health check status is down and fallback method is invoked. MP Metrics are increased
```
fabiobrz: Rewording needed? `Overall and the health check status is down` -> `Overall and health check status are down`.
The same applies to the following test methods Javadoc, too.
WDYT?

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
```diff
@@ -0,0 +1,204 @@
...
+        MetricsChecker.get(metricsRequest)
+                .validateSimulationCounter(FailSafeDummyService.MAX_RETRIES + 1) // 1 call + N retries
+                .validateInvocationsTotal(1)
+                .validateInvocationsFailedTotal(1)
+                .validateRetryRetriesTotal(FailSafeDummyService.MAX_RETRIES) // N retries
+                .validateRetryCallsFailedTotal(1)
+                .validateRetryCallsSucceededTotal(0)
+                .validateFallbackCallsTotal(1);
+
+        get(HealthUrlProvider.healthEndpoint()).then().statusCode(503);
```
fabiobrz: I guess that headers and body validation is skipped for this one because you already did it on the previous call, right? In such case, what about a single line comment here?

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIHealthBaseTest.java
```diff
@@ -24,23 +24,38 @@
...
-import io.restassured.response.ValidatableResponse;
 import io.restassured.specification.RequestSpecification;
 
 /**
- * Testclass to test {@code /health}, {@code /health/live}, {@code /health/ready} endpoints. Liveness and readiness probes
- * are configured via MP Config properties at the beginning. Tests in the testclass are not expected to change state of
- * probes after the initialization in the beginning since some ways of MP Config configurations in order to change values
- * requires a WildFly to be reloaded.
+ * Testclass to test {@code /health}, {@code /health/live}, {@code /health/ready} endpoints.
+ * {@link CDIBasedLivenessHealthCheck} and {@link CDIBasedReadinessHealthCheck} probes are based on fail-sage CDI bean
```
fabiobrz: Typo: `sage` -> `safe`

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIHealthBaseTest.java
```diff
@@ -24,23 +24,38 @@
...
+ * {@link FailSafeDummyService#IN_MAINTENANCE_CONFIG_PROPERTY} is {@code false}</li>
+ * <li>{@link FailSafeDummyService#READY_IN_MAINTENANCE_CONFIG_PROPERTY} - service is ready if
+ * {@link FailSafeDummyService#IN_MAINTENANCE_CONFIG_PROPERTY} is {@code true}</li>
+ * </ul>
+ * If service is in maintenance a method simulating opening resources {@link FailSafeDummyService#simulateOpeningResources}
+ * throws and {@link IOException}.
+ * Subclasses are supposed to configure only MP Config - set MP Config properties in different ways.
+ *
+ * Tests in the testclass are not expected to change state of probes after the initialization in the beginning since some
+ * ways of MP Config configurations in order to change values requires a WildFly to be reloaded.
```
fabiobrz: I'd reword the final part of this sentence, for instance:
```
Tests in this test class are not expected to change the state of probes after their initialization since some
 ways of configuring MP Config would require a Wildfly reload in order to update those values.
```

But most importantly, I'd like to know whether this affects the logic of the tests that this class is executing. In this case, I'd emphasize this aspect.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIHealthBaseTest.java
```diff
@@ -153,7 +180,7 @@ public void testReadinessEndpointUp() throws Exception {
      */
     @Ignore("WFLY-12924, WFLY-12925")
     @Test
-    public void testHealthEndpointDownInMaintenace() throws Exception {
+    final public void testHealthEndpointDownInMaintenace() throws Exception {
```
fabiobrz: Typo: `..InMaintenace` -> `..InMaintenance`, the same applies for the following methods.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIHealthDynamicBaseTest.java
```diff
@@ -21,7 +21,7 @@
 /**
  * Testclass to test {@code /health}, {@code /health/live}, {@code /health/ready} endpoints. Liveness and readiness probes
  * are configured via MP Config properties at the beginning and changed later in test execution. Tests in the testclass
- * expect no reload operation after changing the values.
+ * expect no reload operation after changing the values (during the test execution).
```
fabiobrz: Here is a better version - IMHO - than the version I commented on previously for `FailSafeCDIHealthBaseTest` class Javadoc (last sentence).

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
istraka: Exatly. I will add one.

istraka:
https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/eap-7.x-microprofile-testsuite/51/ is pending, Local run:
```
10:09:05,084 INFO  [org.jboss.as] (MSC service thread 1-2) WFLYSRV0050: WildFly Full 19.0.0.Beta1-SNAPSHOT (WildFly Core 11.0.0.Beta4) stopped in 53ms
[INFO] 
[INFO] Results:
[INFO] 
[WARNING] Tests run: 61, Failures: 0, Errors: 0, Skipped: 18
```

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
```diff
@@ -0,0 +1,206 @@
...
+        Map<String, String> readyCheck = Collections.unmodifiableMap(new HashMap<String, String>() {
+            {
+                put("name", "dummyReadiness");
+                put("status", "DOWN");
+            }
+        });
+        get(HealthUrlProvider.healthEndpoint()).then()
+                .statusCode(503)
+                .contentType(ContentType.JSON)
+                .header("Content-Type", containsString("application/json"))
```
fabiobrz: Aren't
```
contentType(ContentType.JSON)
```
and
```
header("Content-Type", containsString("application/json"))
```
a different way to check for the same header value? Also in following test methods' checks. Just checking here.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIHealthBaseTest.java
```diff
@@ -57,17 +72,29 @@ public void before() throws ConfigurationException, InterruptedException, Timeou
...
     /**
      * @tpTestDetails Multiple-component customer scenario - health check status is based on fail-safe CDI bean and MP Config
      *                property.
-     * @tpPassCrit Overall and the health check status is up. MP Metrics are increased according to the specification.
+     * @tpPassCrit Overall and the health check status are up. MP Metrics are increased according to the specification.
      * @tpSince EAP 7.4.0.CD19
      */
     @Test
-    public void testHealthEndpointUp() throws Exception {
+    final public void testHealthEndpointUp() throws Exception {
```
fabiobrz: Few lines below I can see duplicate `.contentType` checks.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
```diff
@@ -0,0 +1,206 @@
...
+    @ArquillianResource
+    protected ManagementClient managementClient;
+
+    @Before
+    public void before() throws ConfigurationException, InterruptedException, TimeoutException, IOException {
+        ArquillianContainerProperties arqProps = new ArquillianContainerProperties(
+                ArquillianDescriptorWrapper.getArquillianDescriptor());
+        String url = "http://" + arqProps.getDefaultManagementAddress() + ":" + arqProps.getDefaultManagementPort()
+                + "/metrics";
+        metricsRequest = given().baseUri(url).accept(ContentType.JSON);
```
fabiobrz: I was wondering whether it is mandatory that what is above - i.e. `metricsRequest` configuration is done before _each_ test. Wouldn't `@BeforeClass` be enough? Also with some concern about the order of extended `@Before` methods execution.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIHealthBaseTest.java
```diff
@@ -57,17 +72,29 @@ public void before() throws ConfigurationException, InterruptedException, Timeou
         new Administration(ManagementClientProvider.onlineStandalone(managementClient)).reload();
```
fabiobrz: See my comment above, about evaluating `@BeforeClass` for this one too.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
istraka: Yeah, it is the same...I guess it has came from Throntail TS, will fix...

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
istraka: The class no longer extends anything. I will move it to `@BeforeClass`

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIHealthBaseTest.java
istraka: This also resets MP Metrics...will change the comment. This must be here.

istraka:
Fixed
```
[INFO] Results:
[INFO] 
[WARNING] Tests run: 63, Failures: 0, Errors: 0, Skipped: 17
```

istraka:
Add fix for https://github.com/jboss-eap-qe/eap-microprofile-test-suite/issues/162
```
[INFO] Results:
[INFO] 
[WARNING] Tests run: 63, Failures: 0, Errors: 0, Skipped: 17
```

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
fabiobrz: So why is `reload` needed _before each method execution, now? Just asking, I am aware I might be missing something.

istraka: Some tests check MP Metrics. Without reload the metric from a test would be preserved and have an impact on other tests.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIHealthBaseTest.java
fabiobrz: I see, so that is meant to reset Metrics _before_each test, right? And this applies to the above comment as well, I guess. 
I am asking because I was concerned about the order of execution for `@Before` methods in the hierarchy (see https://junit.org/junit4/javadoc/latest/org/junit/Before.html - superclass comes _first_) so this is good if the goal is _just_ to reset metrics.
The metrics URL could be computed just once here too, IMO, WDYT?

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIConfigFileHealthTest.java
fabiobrz: Thanks, see my comment below.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIHealthBaseTest.java
istraka: You are right, I must have missed this one.

istraka:
Fixed
```
[INFO] Results:
[INFO] 
[WARNING] Tests run: 63, Failures: 0, Errors: 0, Skipped: 17
```

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIModelFilePropsHealthTest.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+@RunWith(Arquillian.class)
+@RunAsClient
+@ServerSetup({ MicroProfileFTSetupTask.class, FailSafeCDIModelFilePropsHealthTest.SetupTask.class })
+public class FailSafeCDIModelFilePropsHealthTest extends FailSafeCDIHealthBaseTest {
+
+    private byte[] liveBytes;
+    private byte[] readyBytes;
+    private byte[] inMaintenanceBytes;
+    private byte[] readyinMaintenanceBytes;
```
fabiobrz: Isn't this one supposed to be `readyInMaintenanceBytes` (capital "i" in "In")? I see you're using this notation basically everywhere - e.g. as in `readyInMaintenanceFile`.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/FailSafeCDIModelFilePropsHealthTest.java
```diff
@@ -0,0 +1,100 @@
...
+    Path readyInMaintenanceFile = Paths.get(
+            FailSafeCDIModelFilePropsHealthTest.class
+                    .getResource("file-props/" + FailSafeDummyService.READY_IN_MAINTENANCE_CONFIG_PROPERTY).getPath());
+
+    @Before
+    public void backup() throws IOException {
+        liveBytes = Files.readAllBytes(liveFile);
+        readyBytes = Files.readAllBytes(readyFile);
+        inMaintenanceBytes = Files.readAllBytes(inMaintenanceFile);
+        readyinMaintenanceBytes = Files.readAllBytes(inMaintenanceFile);
```
fabiobrz: I think that `readyInMaintenanceFile` should be read, here.

istraka:
Fixed
```
[INFO] Results:
[INFO] 
[WARNING] Tests run: 63, Failures: 0, Errors: 0, Skipped: 17
```

istraka:
```
[INFO] Results:
[INFO] 
[WARNING] Tests run: 63, Failures: 0, Errors: 0, Skipped: 17
```

istraka:
There is an intermittent issue with creaper wrapping arquillian client
The concept is wrong and we can not wrap the client anymore
https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/eap-7.x-microprofile-testsuite/jdk=oracle-java-11,label_exp=RHEL7/lastCompletedBuild/testReport/
```
[INFO] Results:
[INFO] 
[WARNING] Tests run: 63, Failures: 0, Errors: 0, Skipped: 17
```

fabiobrz:
> There is an intermittent issue with creaper wrapping arquillian client
> The concept is wrong and we can not wrap the client anymore
> https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/eap-7.x-microprofile-testsuite/jdk=oracle-java-11,label_exp=RHEL7/lastCompletedBuild/testReport/
> 
> ```
> [INFO] Results:
> [INFO] 
> [WARNING] Tests run: 63, Failures: 0, Errors: 0, Skipped: 17
> ```

Hi @istraka, as agreed, let's open an issue and put the PR on hold - as this is affecting all the MP specs modules in the TS.

istraka:
On hold: https://github.com/jboss-eap-qe/eap-microprofile-test-suite/issues/170

istraka:
Rebased & fixed. Please review.
```
[INFO] Results:
[INFO] 
[WARNING] Tests run: 63, Failures: 0, Errors: 0, Skipped: 17
```

honza-kasik:
LGTM! Thank you @istraka - I appreciate expanded comments! Merging...

PR #198, opened 2020-01-23
WIP

istraka:
- [ ] Pull Request contains a description of the changes
- [ ] Pull Request does not include fixes for multiple issues/topics
- [ ] Code is formatted, imports ordered, code compiles and tests are passing
- [ ] Link to the passing job is provided
- [ ] Code is self-descriptive and/or documented
- [ ] Description of the tests scenarios is included (see #46)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ jdk:
 install:
  - mvn -version; docker version
 script:
+ - [[ $(git grep import | grep '\*' | grep -v '.mvn' | wc -l) == 0 ]] || ( echo "Wildcard found!" && exit 1 )
  - mvn -B formatter:validate impsort:check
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,6 +5,7 @@ jdk:
 install:
  - mvn -version; docker version
 script:
+ - [[ $(git grep import | grep '\*' | grep -v '.mvn' | wc -l) == 0 ]]
```
honza-kasik: I know this is a WIP and so on but would this handle an echo?

`... == 0 || echo "Wildcard import found!"`

istraka: IMO you can use this (without `[[ ]]`) when you test a variable. I need to execute bash command. However good point with echo a message. That is definitely needed!

PR #200, opened 2020-01-23
Removing throws Exception. Fixes #194

### LINKED ISSUE #194: Methods in `org.jboss.eap.qe.ts.common.docker.Docker` tool mask exceptions
Almost every method in `org.jboss.eap.qe.ts.common.docker.Docker` has `throws Exception` in its signature which causes masking of exception and leaves no precise information what kind of exception method is throwing.

honza-kasik:
* Not using `throws Exception`
* Not using custom unchecked exception
* Using unchecked exceptions as little as possible
* Present a public interface for possible future extension (podman?)
* Introducing several checked exceptions

- [ ] Link to the passing job is provided - 3 tests are failing, doesn't seem to be caused by this PR https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/eap-7.x-microprofile-testsuite/jdk=oracle-java-11,label_exp=RHEL7/93/


tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -67,85 +128,83 @@ public void start() throws Exception {
...
-        Process dockerRunProcess = new ProcessBuilder()
-                .redirectErrorStream(true)
-                .command(new String[] { "docker", "ps" })
-                .start();
-
-        dockerRunProcess.waitFor();
+    public boolean isRunning() {
+        Process dockerRunProcess;
+        try {
+            dockerRunProcess = new ProcessBuilder()
```
mnovak1: I think that `runDockerCommand()` might be used here as well.

tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -67,85 +128,83 @@ public void start() throws Exception {
...
-        dockerInfoProcess.waitFor();
-        if (dockerInfoProcess.exitValue() != 0) {
-            throw new DockerException("Docker is either not present or not installed on this machine. It must be installed " +
-                    "and started up for executing tests with docker container.");
+    private boolean isDockerPresent() {
+        try {
+            Process dockerInfoProcess = new ProcessBuilder()
+                    .redirectErrorStream(true)
+                    .command(new String[] { "docker", "info" })
+                    .start();
```
mnovak1: I think that `runDockerCommand()` might be used here as well.

mnovak1:
@honza-kasik do you think that you might take a look at the test failures, I think they might be related to this PR

honza-kasik:
@mnovak1 Exactly. I removed this calls because it didn't make sense to me. The problem is that the  docker start command actually creates and starts container. Adding calls to these back to methods is IMHO a bad idea. I'll try to solve this...

honza-kasik:
Hopefully all fixed. Job is running https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/eap-7.x-microprofile-testsuite/98/

honza-kasik:
@mnovak1 All green

tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -23,27 +26,85 @@
...
+            dockerRunProcess = new ProcessBuilder()
+                    .redirectErrorStream(true)
+                    .command(dockerStartCommand)
+                    .start();
+        } catch (IOException e) {
+            throw new ContainerStartException("Failed to start the '" + String.join(" ", dockerStartCommand) + "' command");
+        }
 
-    public void start() throws Exception {
+        this.outputPrintingThread = startPrinterThread(dockerRunProcess, this.out, this.name);
```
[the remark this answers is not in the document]
mnovak1: what do you think about this one?

tooling-docker/src/main/java/org/jboss/eap/qe/ts/common/docker/Docker.java
```diff
@@ -156,54 +259,104 @@ public boolean isRunning() throws Exception {
...
+        cmd.add(dockerCommand);
+        Collections.addAll(cmd, commandArguments);
+        try {
+            final Process process;
+
+            process = new ProcessBuilder()
+                    .command(cmd)
+                    .start();
+
+            process.waitFor(10, TimeUnit.SECONDS);
```
mnovak1: I think this should be removed here and leave to handle `Process` on calling method.

microprofile-fault-tolerance/src/test/java/org/jboss/eap/qe/microprofile/fault/tolerance/DatabaseCrashTest.java
```diff
@@ -205,6 +205,7 @@ public void dropDatabaseSchema() {
     public static void tearDown() throws Exception {
         // stop DB if still running and delete data directory
         postgresDB.stop();
+        postgresDB.remove();
```
mnovak1: I think `remove()` should be called from stop() and kill() medhod...as this might be source of failures.

honza-kasik: The issue I see here is that the method called `stop` removes a container. That is IMHO not desired. What I have been thinking about is implementing `Closable/AutoClosable` interface to have a `close` method. Which would care about cleanup when the container is no longer needed. WDYT?

mnovak1: I've found that there can be used `--rm` option in `docker run...` which causes that container is automatically removed when stopped (or killed) what about to use it and get rid of `remove();`

honza-kasik:
@mnovak1 Thanks! Updated https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/eap-7.x-microprofile-testsuite/114/

PR #206, opened 2020-02-20
MP health disable default procedures

fabiobrz:
* Fixing "old" tests that were not providing a readiness check in order to set `mp.health.disable-default-procedures` to true so that default readiness checks are not created and assertions are working (after https://issues.redhat.com/browse/WFLY-12952 and https://github.com/wildfly/wildfly/pull/12940).

* Adding test coverage to assess basic behavior of default readiness probe result when mp.health.disable-default-procedures property is set to false (default), see https://download.eclipse.org/microprofile/microprofile-health-2.2/microprofile-health-spec.html#_disabling_default_vendor_procedures and https://github.com/wildfly/wildfly/pull/12940

- N/A - Link to the passing job is provided: Jenkins is super busy, tests pass locally

### DIFF
--- a/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/HealthDeprecatedTest.java
+++ b/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/HealthDeprecatedTest.java
@@ -14,6 +14,7 @@ import org.jboss.eap.qe.microprofile.tooling.server.configuration.ConfigurationE
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.EmptyAsset;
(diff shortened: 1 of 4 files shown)


microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/HealthDeprecatedTest.java
```diff
@@ -28,6 +29,8 @@
     public static Archive<?> deployment() {
         return ShrinkWrap.create(WebArchive.class, HealthDeprecatedTest.class.getSimpleName() + ".war")
                 .addClasses(DeprecatedHealthCheck.class)
+                .addAsManifestResource(new StringAsset("mp.health.disable-default-procedures=true"),
+                        "microprofile-config.properties")
```
marekkopecky: I like this approach, it is aligned with spec directly.

PR #221, opened 2020-09-07
WIP: Move CustomConfig* tests to the separate Maven test execution

zhantemirov:
Hi @istraka 

After some MP Config studying and Metrics failures [1] investigation I tried to fix those failures, though I'm not quite sure whether this is the correct way.

So the main issue was the unset `config.source.properties.path` system property. I added the existing SetupTask class to the ServerSetup annotation to all failing test cases.

I'd like to hear your opinions regarding the solution (and maybe another ones) from you, guys - @istraka @fabiobrz, thanks!

[1] -https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/eap-7.x-microprofile-testsuite/jdk=oracle-java-11,label_exp=RHEL7/353/#showFailuresLink

Passing job - https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/eap-7.x-microprofile-testsuite/jdk=oracle-java-11,label_exp=RHEL7/355/

- [X] Pull Request contains a description of the changes
- [X] Pull Request does not include fixes for multiple issues/topics
- [X] Code is formatted, imports ordered, code compiles and tests are passing
- [X] Link to the passing job is provided
- [X] Code is self-descriptive and/or documented
- [X] Description of the tests scenarios is included (see #46)

### DIFF
--- a/microprofile-metrics/pom.xml
+++ b/microprofile-metrics/pom.xml
@@ -48,4 +48,53 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+
+    <profiles>
+        <profile>
+            <id>metrics-default</id>
+            <activation>
+                <property>
+                    <name>!skip-metrics-default</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-surefire-plugin</artifactId>
+                    <version>${version.org.apache.maven.plugins.maven-surefire-plugin}</version>
+                    <executions>
+                        <execution>
+                            <id>default-test</id>
+                            <phase>test</phase>
+                            <goals>
+                                <goal>test</goal>
+                            </goals>
+                            <configuration>
+                                <excludes>
+                                    <exclude>org/jboss/eap/qe/microprofile/metrics/integration/config/CustomMetricCustomConfigSourceTest.java</exclude>
+                                    <exclude>org/jboss/eap/qe/microprofile/metrics/integration/config/CustomMetricCustomConfigSourceProviderTest.java</exclude>
+                                </excludes>
+                            </configuration>
+                        </execution>
+                        <execution>
+                            <id>two-test</id>
+                            <phase>integration-test</phase>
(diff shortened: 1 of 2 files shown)


istraka:
`config.source.properties.path` property is supposed to be used by `CustomConfigSource` which is added as module. Its purpose is to load MP Config properties from file dynamically and provide them for MP Config itself. The property should not be needed anywhere else. IMO using `CustomConfigSource` setup class for tests that aren't designed to use it is a not a good idea.

```
So the main issue was the unset config.source.properties.path system property
```
I see the change in `CustomMetricCustomConfigSourceTest.SetupTask#tearDown()` where removing the property was missing. Yes, the setupTask should remove the property. But the presence should matter only in `CustomMetricCustomConfigSourceTest` and `CustomMetricCustomConfigSourceProviderTest`

The exception comes from https://github.com/jboss-eap-qe/eap-microprofile-test-suite/blob/master/microprofile-health/src/main/java/org/jboss/eap/qe/microprofile/health/integration/CustomConfigSource.java#L23

This class should be used only with that two test classes as a module. Can you look into why the module is presented?

istraka (changes requested):
See the comment.

zhantemirov:
@istraka thanks a lot for your input. 

I did some additional analysis and found out the tests are passing when they are not run with _CustomMetricCustomConfigSource*_ tests (hereinafter _Tests_). Obviously there is an issue with module configuration (adding/removing). 

But due the lack of time and my experience in MP Config I ended up with a different approach - moving _Tests_ to the separate Maven test step - in other words, all tests besides the _Tests_ are run in the first phase, while the _Tests_ are run in the second phase.

Passing job link is in the PR description.

istraka:
I see. This is definitely a better workaround than the one before. However this would bring another complexity to the TS and the issue would be still there. We should either fix the issue or be very sure this is the only option available. Feel free to reach me out on a gchat to discuss the technical details.

zhantemirov:
PR does not make sense anymore due to the latest commentaries in the reported JIRA - https://issues.redhat.com/browse/WFLY-13779

PR #222, opened 2020-09-08
Enable TS to run tests against bootable JAR

fabiobrz:
This PR is to allow the TS to run tests against the bootable JAR and adds profiles to the existing modules in order to do so.

Successful job run: https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/eap-7.x-microprofile-testsuite-bootable/66

- N/A - Description of the tests scenarios is included (see #46)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,6 @@
         <testsuite.galleon.pack.version>${env.WILDFLY_GALLEON_PACK_VERSION}</testsuite.galleon.pack.version>
         <!-- Bootable JAR -->
         <version.org.wildfly.jar.plugin>${env.BOOTABLE_JAR_MAVEN_PLUGIN_VERSION}</version.org.wildfly.jar.plugin>
-        <!--        <version.org.wildfly.arquillian.bootable>3.0.0.Beta1-SNAPSHOT</version.org.wildfly.arquillian.bootable>-->
     </properties>
 
     <dependencyManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
     <packaging>pom</packaging>
 
     <name>Test suite for MicroProfile on WF/EAP</name>
-    
+
     <modules>
         <module>tooling-cpu-load</module>
         <module>tooling-docker</module>
@@ -221,7 +221,7 @@
         <connection>scm:git:git@github.com:jboss-eap-qe/eap-microprofile-test-suite.git</connection>
         <developerConnection>scm:git:git@github.com:jboss-eap-qe/eap-microprofile-test-suite.git</developerConnection>
     </scm>
-    
+
     <build>
         <plugins>
             <plugin>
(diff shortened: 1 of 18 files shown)


pom.xml
```diff
@@ -14,7 +14,7 @@
     <packaging>pom</packaging>
 
     <name>Test suite for MicroProfile on WF/EAP</name>
-    
+
     <modules>
         <module>tooling-cpu-load</module>
         <module>tooling-docker</module>
```
marekkopecky, line 20: What about `tooling-docker`? I don't see any updates for this module. Are tests from this module executed in a same way like without bootable profile?

pom.xml
```diff
@@ -484,5 +502,80 @@
...
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+            <dependencyManagement>
+                <dependencies>
+                    <!-- Bootable JAR related dependencies -->
+                    <dependency>
+                        <groupId>org.wildfly.arquillian</groupId>
+                        <artifactId>wildfly-arquillian-container-bootable</artifactId>
```
marekkopecky, line 556: What about to add this to dependencies directly (not to dependencyManagement)? It seems that this dependency is used in all modules relevant modules. In this case, this dependency don't need to be defined in each MP sub-modules in bootable profile.

pom.xml
```diff
@@ -484,5 +502,80 @@
...
+                            <!-- superseded by jakarta.inject:jakarta.inject-api -->
+                            <exclusion>
+                                <groupId>javax.inject</groupId>
+                                <artifactId>javax.inject</artifactId>
+                            </exclusion>
+                        </exclusions>
+                    </dependency>
+                    <dependency>
+                        <groupId>org.wildfly.core</groupId>
+                        <artifactId>wildfly-launcher</artifactId>
```
marekkopecky, line 573: Btw, what is this dependency about? It is used in 4 modules out of 7. Can it be used for all 7 (although not needed for all of them) and be defined directly in main pom.xml (to decrease maintenance for 4 same definitions in 4 different places)?

marekkopecky:
@fabiobrz Thank you for this PR, I added a few minor comments ...

pom.xml
fabiobrz: I am not against excluding something but this module is not tested like the other ones (involving Arquillian) but just two Wildfly containers.
Indeed, this is the only module between the `tooling-*` ones which didn't require for the testing to be adapted in order to prevent problems with Arquillian. 
Basically the `tooling-*` modules are tested only with the standard TS execution.

pom.xml
fabiobrz: Eh, as said above this would me for this to be shaded into the `tooling-*` modules as well, although those are not tested (and IMO shouldn't be) when executing the Bootable JAR profile. WDYT?

pom.xml
fabiobrz: It is the API for WIldFly launch. For instance a new command builder to start WildFly has been introduced in order to support the execution in case of bootable JAR. This is needed by the WildFly Arquillian bootable container adapter.

pom.xml
honza-kasik: @fabiobrz The dependencies are specified per-profile. Wouldn't this mean they are needed only when the profile is activated?

fabiobrz: Thanks @honza-kasik - I now get what you and @marekkopecky mean. Iìve fixed it, let me know what you think.

honza-kasik: @fabiobrz Looks good, thank you! I guess it works as usual too, correct?

pom.xml
honza-kasik: tooling-docker doesn't contain any arq tests AFAIK, resolving, thank you.

pom.xml
honza-kasik: Seems good to me as it is currently - using just a reference from dependency management.

pom.xml
fabiobrz: @honza-kasik - looks like it's working, yep, see https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/eap-7.x-microprofile-testsuite-bootable/67/

Issue #215, opened 2020-04-20
Evaluate MP TS branching with respect to expected EAP expansion pack


mnovak1:
EAP 7.3.x.CP will be delivered with XP 1.x but with release of MP 4.0 (planned in summer 2020) there will be XP 2.x which will co-exist with XP 1.x for some time and incompatible with XP 2.x. Thus it's not possible to have `7.3.x` branch. 

Suggestion is to branch MP TS by XP versions like:
- `7.3-xp-1.x` and `7.3-xp-2.x`
- `7.4-xp-2.x and so on

PR #223, opened 2020-09-11
[WFLY-12342] Add server probes for readiness health check

mchoma:
[DO NOT MERGE] This needs [WF21 changes in](https://github.com/wildfly/wildfly/pull/13397) to be valid. 

Changes relater to WFLY-12342:
- adjust existing tests to work with new changes
- added new tests for new fetures in WFLY-12342

CI run: :heavy_check_mark:  https://eap-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/eap-7.x-microprofile-custom-build-from-sources-face/120/
(metrics tests are expected to fail currently with 21 and are tracked separately by https://issues.redhat.com/browse/WFLY-13779)

### DIFF
--- a/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/delay/ReadinessChecker.java
+++ b/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/delay/ReadinessChecker.java
@@ -51,20 +51,20 @@ public class ReadinessChecker implements Callable<Boolean> {
                         if (response.getStatusCode() == 503) {
                             if (checks == null) {
                                 addState(DOWN_NO_CONTENT());
-                            } else if (checks.size() == 0) {
+                            } else if ((checks.size() == 4) && contains(checks, "empty-readiness-checks")) {
                                 addState(DOWN_NO_CHECK());
-                            } else if (checks.get(0).get("name").equals(DelayedReadinessHealthCheck.NAME)) {
+                            } else if ((checks.size() == 4) && contains(checks, DelayedReadinessHealthCheck.NAME)) {
                                 addState(ReadinessState.DOWN_WITH_CHECK());
                             }
                         } else if (response.getStatusCode() == 200) {
                             if (checks == null) {
                                 throw new RuntimeException("Readiness probe is UP (200) however missing JSON content");
                             }
-                            if (checks.size() == 0) {
+                            if ((checks.size() == 4) && contains(checks, "empty-readiness-checks")) {
                                 addState(UP_NO_CHECK());
-                            } else if (checks.get(0).get("name").equals(DelayedReadinessHealthCheck.NAME)) {
+                            } else if ((checks.size() == 4) && contains(checks, DelayedReadinessHealthCheck.NAME)) {
                                 addState(UP_WITH_CHECK());
-                            } else if (checks.get(0).get("name").startsWith(DEFAULT_READINESS_CHECK_NAME_PREFIX)) {
+                            } else if ((checks.size() == 4) && contains(checks, DEFAULT_READINESS_CHECK_NAME_PREFIX)) {
                                 addState(UP_WITH_DEFAULT_CHECK());
                             }
                         }
@@ -90,6 +90,23 @@ public class ReadinessChecker implements Callable<Boolean> {
 
     private int lastIndexOfStates() {
         return states.size() - 1;
+    }
+
+    /**
+     * @param list collection to find in
+     * @param key which we are looking for
+     * @return true if key {@code name} atribute startsWith {@code key}. False otherwise.
+     */
+    private boolean contains(List<Map<String, String>> list, String key) {
+        if (list == null) {
+            return false;
+        }
+        for (Map<String, String> item : list) {
+            if (item.get("name").startsWith(key)) {
+                return true;
+            }
+        }
+        return false;
     }
 
     public void stop() {
--- a/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/DefaultReadinessProcedureHealthTest.java
+++ b/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/DefaultReadinessProcedureHealthTest.java
@@ -5,19 +5,29 @@ import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.hasItems;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
+import static org.jboss.eap.qe.microprofile.tooling.server.configuration.creaper.ManagementClientHelper.executeCliCommand;
 
+import java.io.IOException;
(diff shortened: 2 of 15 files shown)


mchoma:
I would appreciate if someone with more experience in this testsuite area will have a look. @istraka you was suggested by github, is there chance you can have a look? :)

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/delay/ReadinessChecker.java
```diff
@@ -51,20 +51,20 @@ public Boolean call() {
                         if (response.getStatusCode() == 503) {
                             if (checks == null) {
                                 addState(DOWN_NO_CONTENT());
-                            } else if (checks.size() == 0) {
+                            } else if ((checks.size() == 4) && contains(checks, "empty-readiness-checks")) {
                                 addState(DOWN_NO_CHECK());
-                            } else if (checks.get(0).get("name").equals(DelayedReadinessHealthCheck.NAME)) {
+                            } else if ((checks.size() == 4) && contains(checks, DelayedReadinessHealthCheck.NAME)) {
```
istraka, line 56: Can we verify new probes here? running test with parallel (separate terminal) `while true; do echo $(date +"%T.%3N") ;  curl   localhost:19990/health/ready; echo ""; done` produces:
```
...
===>>server does not exists
17:23:48.141
curl: (7) Failed to connect to localhost port 19990: Connection refused
===>>server is starting
17:23:48.154
{"status":"DOWN","checks":[{"name":"deployments-status","status":"DOWN"},{"name":"empty-readiness-checks","status":"DOWN"},{"name":"boot-errors","status":"DOWN"},{"name":"server-state","status":"DOWN"}]}
...
===>>deployment readiness probes are installed, deployment not ready
17:23:56.073
{"status":"DOWN","checks":[{"name":"deployments-status","status":"DOWN"},{"name":"boot-errors","status":"DOWN"},{"name":"server-state","status":"DOWN"},{"name":"delayed-readiness","status":"DOWN"}]}
===>>deployment readiness probes are installed, deployment ready
17:23:59.094
{"status":"DOWN","checks":[{"name":"deployments-status","status":"UP","data":{"NotReadyHealthCheckTest.war":"OK"}},{"name":"boot-errors","status":"UP"},{"name":"server-state","status":"UP","data":{"value":"running"}},{"name":"delayed-readiness","status":"DOWN"}]}
```

IMO we can increase test coverage here. Even for `UP` states.

But I don't know test plan and so on...So not sure if it not out of scope of the PR.

istraka (changes requested):
One unnecessary change. 
One open suggestion?

Otherwise LGTM.

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/delay/ReadinessChecker.java
mchoma: Sorry I do not get your comment. fully. Where does `===>>` comes from? Do you see some problem while doing manual tests?

I must admit understanding and changing this class was for me tricky. Maybe we can arrange short call to sync.

istraka: As discussed `===>>` is added by me to describe server state at the moment.

The main purpose of the test case is to verify that server is not ready (`UP`) before deployment probe is installed and would return `DOWN` from the begging. The scenario is still valid and the RFE does not change it. After discussion we have agreed that test coverage in `DefaultReadinessProcedureHealthTest` is sufficient. 

Thus the discussion has been resolved.

mchoma:
Question is when to merge it? Wildfly 21.0.Beta1 is released, so there is already publicly available distro this testsuite will work with. 

But that can potentially break running 7.3.x releases. For EAP XP1 there is branch XP1. Have you already have such situation in past?

istraka (approved):

istraka:
@mchoma I have been out of this project for a while so I do not know. IMO if `7.3.z` used `XP1` branch it could create a confusion. What about creating `7.3.z` branch? WDYT @fabiobrz @honza-kasik?

mchoma:
I agree it is confusing. But XP1 and 7.3.x have common Observability specs so I would say it is correct from feature perspective. We can rename it to something what fits both. My motivation is to reduce number of necessary branches to avoid cherry-picking. But of course branch per stream is most correct. But when you will have fix in observability specs tests you will need to make sure they are in all 3 branches master, 7.3.x and XP1. Thats tradeoff.

mchoma:
Actually MP health changes will be present in EAP 7.3.4 Base as well. That mean testsuite `master` branch will be still valid for testing EAP 7.3.4 Base.

So once EAP 7.3.3 Base will be tested we can merge this. We will need to cherry pick to XP1 branch as well.

mchoma:
I believe as EAP 7.3.3 Base was released this is safe to merge now into master.

mnovak1:
I think this is ok to merge as master should work with latest WF. 

I guess now we're facing question whether to use branches for EAP 7.3.z/XP1,2,3 or have less branches and use profiles to run separate groups of tests. For example XP 3 will contain MP 4.0 where might be other incompatible changes so tests for those should not be executed with XP1,2 and so on. 
I don't have strong preference here. I'm used to profiles as cherry-picking to different branches makes it harder track what is where.

mchoma:
@mnovak1 Thank you for review. Can you merge it please, then?

I have put on meeting agenda how we should manage XP3/ MP4. If TT comes to best is module/per spec we should probably copy that approach. We should double check if they are satisfied with this model. I see some drawback as well.

fabiobrz:
> I think this is ok to merge as master should work with latest WF.
> 
> I guess now we're facing question whether to use branches for EAP 7.3.z/XP1,2,3 or have less branches and use profiles to run separate groups of tests. For example XP 3 will contain MP 4.0 where might be other incompatible changes so tests for those should not be executed with XP1,2 and so on.
> I don't have strong preference here. I'm used to profiles as cherry-picking to different branches makes it harder track what is where.
> 
> Edit: Now i'm thinking that there are MP specs directly included in modules thus to avoid version conflicts we would need to split those modules per MP spec version as it's done in https://github.com/rhoar-qe/thorntail-test-suite/tree/master/microprofile

Thanks @mnovak1 - so, following up suggestions from you and Martin, IIUC we should implement packages vs. spec version here, right? Would this concretely mean a major refactoring of the TS is needed?

mnovak1:
@fabiobrz @mchoma I've found myself to create issue for this a half year ago :-) - https://github.com/jboss-eap-qe/eap-microprofile-test-suite/issues/215. I've added there suggestion with profiles. Feel free to edit or comment it so we can reach some consensus.

mchoma:
@mnovak1 Can you merge as we agree this is safe to be merged into master now. We are waiting for this changes to be present for XP2 pipeline.

I found discussion about organisation of testsuite unrelated. But for sure we need to continue with its discussion on issue you linked. Changes of this MR are not related to MP spec upgrade. We just added vendor specific probes to existing MP Health Spec.

mnovak1 (approved):

mnovak1:
Done.

PR #231, opened 2021-01-29
Allow to use podman via system property even without docker bridge

jbliznak:
If you don't want/can't install docker-podman bridge and you only want to use podman you can now do so by leveraging `-Ddocker.command=podman` property.

Regression run without the property: 
* https://master-jenkins-csb-eap-qe.cloud.paas.psi.redhat.com/job/eap-7.x-microprofile-testsuite/519/
* https://master-jenkins-csb-eap-qe.cloud.paas.psi.redhat.com/job/eap-7.x-microprofile-testsuite/522/

The run with property set to podman tested on local Fedora with docker-less environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
  
  Maven is not a hard requirement as `mvnw` is delivered with this repository.
 
- Podman with docker command bridge can be used as a Docker alternative.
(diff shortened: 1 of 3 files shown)


jstourac (approved):

jbliznak:
Thanks for review @jstourac

PR #235, opened 2021-02-10
[EAP7-1593] Integrate MP Health 3.0 into EAP

xstefank:
https://issues.redhat.com/browse/EAP7-1593

I guess the CI will run for the PR? Or should I run it somewhere with my branch first?

### DIFF
--- a/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/MetricsChecker.java
+++ b/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/MetricsChecker.java
@@ -34,69 +34,53 @@ public class MetricsChecker {
     }
 
     /**
-     * Validate MP FT metric {@code isReady.retry.retries.total} for {@link FailSafeDummyService#isReady()}
+     * Validate MP FT metric {@code ft.retry.retries.total} for {@link FailSafeDummyService#isReady()}
      */
     public MetricsChecker validateRetryRetriesTotal(int retriesTotal) {
-        response.body("application.'ft." + FailSafeDummyService.class.getName() + ".isReady.retry.retries.total'",
+        response.body("base.'ft.retry.retries.total;method=" + FailSafeDummyService.class.getName() + ".isReady'",
                 equalTo(retriesTotal));
         return this;
     }
 
     /**
-     * Validate MP FT metric {@code isReady.retry.callsFailed.total} for {@link FailSafeDummyService#isReady()}
+     * Validate MP FT metric {@code ft.retry.calls.total} with {@code retryResult=maxRetriesReached}
+     * for {@link FailSafeDummyService#isReady()}
      */
-    public MetricsChecker validateRetryCallsFailedTotal(int callsFailedTotal) {
-        response.body("application.'ft." + FailSafeDummyService.class.getName() + ".isReady.retry.callsFailed.total'",
-                equalTo(callsFailedTotal));
-        return this;
-    }
-
-    /**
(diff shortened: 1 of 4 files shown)


jbliznak:
CI runs `mvn -B verify -DskipTests -Djboss.dist=foo` so we definitely should have some proper run :)

It should be this, running your branch against WF snapshot from MP4 branch https://master-jenkins-csb-eap-qe.cloud.paas.psi.redhat.com/job/eap-7.x-microprofile-testsuite/534/

For the sake of this PR, we are only interested of test results of microprofile-health module

microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/integration/MetricsChecker.java
```diff
@@ -66,8 +66,8 @@ public MetricsChecker validateRetryCallsSucceededTotal(int callsSucceededTotal)
      */
     public MetricsChecker validateRetryCallsSucceededNotTriedTotal(int callsSucceededNotTriedTotal) {
         response.body(
-                "application.'ft." + FailSafeDummyService.class.getName()
-                        + ".isReady.retry.callsSucceededNotRetried.total'",
+                "base.'ft.retry.calls.total;method=" + FailSafeDummyService.class.getName()
```
jbliznak, line 69: Honestly, I can't say I understand this change, but definitely you should update also the javadoc for this method and the other one too.

jbliznak: For any reviewer, the reason of these changes lies at Fault Tolerance changes https://github.com/eclipse/microprofile-fault-tolerance/blob/master/spec/src/main/asciidoc/release_notes.asciidoc#release-notes-for-microprofile-fault-tolerance-30

jbliznak: > In addition, metrics have moved from the application: scope to the base: scope for consistency with other MicroProfile specifications.

Looking at this more.. I don't understand why all other methods here don't require the similar changes too due to the FT changes, this seems wrong

jbliznak: I see, all tests using the other methods are ignored :facepalm: 
@xstefank would you mind to deal with unignoring these tests and updating metrics checks for them too?

xstefank: @jbliznak I will try to figure this out. Thanks for the investigation.

jbliznak:
Tests are passing :heavy_check_mark:  https://master-jenkins-csb-eap-qe.cloud.paas.psi.redhat.com/job/eap-7.x-microprofile-testsuite/536/

However, could you comment on the changes you did, ie. reducing number of checks in these tests? Knowing only a little about MP specs I don't follow. Is that because the metrics changed so much we need to update test logic as a whole?

xstefank:
@jbliznak yes, exactly. MP FT is now exposing way more less metrics than it used to which is the first reason for changing/removing some of the test checks. The second set of big changes came from the previously ignored tests that were just setting wrong expectation on the executions (the setConfigProperties changes which I would be glad if someone could confirm are right). Other than that there are no functional changes. The verifications check all non-zero values for all tests that I changed.

jbliznak (approved):

mchoma:
I have no objections. Seems to me ready to be merged.

To be sure these 6 failures are really not related to this changes, I have triggered testsuite from MP4 branch [1]

[1] https://master-jenkins-csb-eap-qe.cloud.paas.psi.redhat.com/job/eap-7.x-microprofile-testsuite/537/

PR #236, opened 2021-02-17
DatabaseCrashTest: add :Z option when mounting volumes to avoid probl…

rhusar:
…ems with SELinux

See https://www.projectatomic.io/blog/2015/06/using-volumes-with-docker-can-cause-problems-with-selinux/


jbliznak:
CI run https://master-jenkins-csb-eap-qe.cloud.paas.psi.redhat.com/job/eap-7.x-microprofile-testsuite/541/ (no new failures for MP4 branch) :heavy_check_mark: 

I will go ahead and merge this as this change seems safe.

rhusar:
@jbliznak I assume you have docker newer than docker-1.6, right? So in that case should be fine.

PR #284, opened 2024-03-23
Fix KeycloakIntegrationHighLevelScenarioTest

jbliznak:
There were two major issues with this test:

1) with update to Keycloak 24, there is a new default required action attached to each new user - Verify Profile - that prevents from getting token unless additional action is performed on first login.
We are removing this new check, as we don't need it for tests.

2) there was a race condition when docker container was started and tests started to execute while the initial admin user still wasn't created. It turned out, readiness probe at /health/ready isn't enough to say if we are fully booted for tests, so we need some additional time for perform post-boot tasks before we continue with tests.

CI run at eap-8.x-microprofile-testsuite/861

### DIFF
--- a/microprofile-jwt/src/test/java/org/jboss/eap/qe/microprofile/jwt/keycloak/KeycloakIntegrationHighLevelScenarioTest.java
+++ b/microprofile-jwt/src/test/java/org/jboss/eap/qe/microprofile/jwt/keycloak/KeycloakIntegrationHighLevelScenarioTest.java
@@ -57,6 +57,7 @@ public class KeycloakIntegrationHighLevelScenarioTest {
             .withPortMapping(KEYCLOAK_EXPOSED_HTTP_PORT + ":8080")
             .withEnvVar("KEYCLOAK_ADMIN", KEYCLOAK_ADMIN_USERNAME)
             .withEnvVar("KEYCLOAK_ADMIN_PASSWORD", KEYCLOAK_ADMIN_PASSWORD)
+            .withEnvVar("KC_HEALTH_ENABLED", "true")
             .withCmdArg("start-dev")
             .withCmdArg("--log-level=DEBUG")
             .build();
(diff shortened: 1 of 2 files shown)


microprofile-jwt/src/test/java/org/jboss/eap/qe/microprofile/jwt/keycloak/KeycloakIntegrationHighLevelScenarioTest.java
```diff
@@ -115,7 +115,12 @@ private static boolean isContainerReady(int port) {
             HttpURLConnection connection = (HttpURLConnection) url.openConnection();
             connection.setRequestMethod("GET");
 
-            return connection.getResponseCode() == HttpURLConnection.HTTP_OK;
+            boolean ready = connection.getResponseCode() == HttpURLConnection.HTTP_OK;
+            if (ready) {
```
fabiobrz, line 119: I'd like to be sure the code is reflecting the requirement you described in the PR description: here we actually want to wait for some more time _once_ (and even though) the health check has returned a HTTP 200, right?

fabiobrz (commented):
Hi @jbliznak - and thanks for your work here.
I am ok with the changes, but dropped one comment to be sure we're on the same page with the implementation.

microprofile-jwt/src/test/java/org/jboss/eap/qe/microprofile/jwt/keycloak/KeycloakIntegrationHighLevelScenarioTest.java
jbliznak: correct, even when server reports READY, the admin user might still not be created yet. That is also why it only makes sense to wait after healtch check HTTP 200

fabiobrz (approved):
Thank @jbliznak - chances LGTM.
Merging.

PR #285, opened 2024-04-09
[MP OpenAPI] - Fix the Bootable JAR plugin config in order to add the microprofile-rest-client layer

fabiobrz:
Following up on https://issues.redhat.com/browse/JBEAP-26776, this is adding the microprofile-rest-client layer  to the Bootable JAR plugin configuration for the `microprofile-open-api` module.

This seems to be the valid configuration ATM, so we need align it to the EAP XP 5 feature pack definition.

Internal **CI run** reference: 
- **job**: eap-8.x-microprofile-simple-face
- **run**: 103 

Please make sure your PR meets the following requirements:
- [x] Pull Request contains a description of the changes
- [x] Pull Request does not include fixes for multiple issues/topics
- [x] Code is formatted, imports ordered, code compiles and tests are passing
- [x] Link to the passing job is provided
- [x] Code is self-descriptive and/or documented
- **N/A** Description of the tests scenarios is included (see #46)

PR #286, opened 2024-04-10
[MP Fault Tolerance] - Fixing the PostrgreSql image TAG since latest is no longer valid

fabiobrz:
The used image/tag is no longer supported, see https://quay.io/repository/centos7/postgresql-13-centos7, so a quick fix is to use the `:centos7`, while the actual solution would be to upgrade the supported PostgreSql image version, e.g.: https://quay.io/repository/sclorg/postgresql-15-c9s?tab=info

Internal **CI run** reference: 
- **job**: eap-8.x-microprofile-simple-face
- **run**: 104 

**N.B.**: failures in microprofile-open-api are expected and fixed by #285 

- **N/A** Code is self-descriptive and/or documented
- **N/A** Description of the tests scenarios is included (see #46)


jbliznak (approved):
As quick fix this is ok, but as you wrote, we should update the image anyway, please make followup issue for it.

PR #290, opened 2024-07-11
[issue 263] - Removing unneeded setup task class that disables all default procedures

### LINKED ISSUE #263: Disable the health default procedures
Disabling the health default procedures by using the `DisableDefaultHealthProceduresSetupTask` class seems to be way too complicated. The MP Configuration specs mentions that this also could be done by setting the `mp.health.disable-default-procedures` property to true in the `microprofile-config.properties` file.
I tried that but it didn't work. 
Also, I can see that the class `MultiDeploymentHealthTest` tries to disable the health default procedures as follows:

        @Deployment(name = "deployment1", order = 1, testable = false)
        public static Archive<?> deployment1() {
          return ShrinkWrap.create(WebArchive.class, MultiDeploymentHealthTest.class.getSimpleName() + "-1.war")
                ...
                .addAsManifestResource(new StringAsset("mp.health.disable-default-procedures=true"),
                        "microprofile-config.properties")
                ...
        }

However, it still has the following annotation:

    @ServerSetup({ DisableDefaultHealthProceduresSetupTask.class })

which duplicates the same functionality. Commenting out the annotation above, the Arquillian tests are failing, showing that only setting the property in the deployment method isn't enough.

So, I have the following questions:
  - is it possible to disable the health default procedures by setting the associated property in `microprofile-config.properties` and, if yes, why doesn't it work here ?
  - is it possible to do it by setting the associated property in the deployment method, as shown above and, if yes, why doesn't it work here and why did you need to annotate with `@ServerSetup` ?
  - is there any simpler way to do it ?

Many thanks in advance.

Kind regards,

Nicolas

fabiobrz:
Follow up to WFLY-19147.

Fixes #263 

- ~Draft because the upstream fix ended up in WildFly 33.x codebase, so we need to decide whether to branch for XP 5 CPs testing, or wait until it will be backported.~

- **Validation runs reference**:  
1. job: **eap-8.x-microprofile-testsuite** 
2. run nr.: **1074** :heavy_check_mark:  (the MP Health related failures here are unrelated and are due to #298)
 
---
- [x] ~Link~ Reference to the passing job is provided

### DIFF
--- a/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/HealthNullTest.java
+++ b/microprofile-health/src/test/java/org/jboss/eap/qe/microprofile/health/HealthNullTest.java
@@ -8,7 +8,6 @@ import static org.hamcrest.Matchers.is;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.container.test.api.RunAsClient;
 import org.jboss.arquillian.junit.Arquillian;
-import org.jboss.as.arquillian.api.ServerSetup;
 import org.jboss.eap.qe.microprofile.health.tools.HealthUrlProvider;
 import org.jboss.eap.qe.microprofile.tooling.server.configuration.ConfigurationException;
 import org.jboss.eap.qe.microprofile.tooling.server.configuration.deployment.ConfigurationUtil;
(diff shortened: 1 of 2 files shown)


marekkopecky:
Thank you for this PR, the fix corresponds with https://github.com/jboss-eap-qe/eap-microprofile-test-suite/issues/263 issue. CI is little unstable but affected tests hasn't failed there. Merging.

PR #313, opened 2024-11-12
Update MicroProfile OpenAPI tests for v4.0.

pferraro:
Parameterized test artifacts for future ease of future upgrades.

### DIFF
--- a/micrometer/pom.xml
+++ b/micrometer/pom.xml
@@ -13,10 +13,26 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.eclipse.microprofile</groupId>
-            <artifactId>microprofile</artifactId>
-            <type>pom</type>
+            <groupId>jakarta.annotation</groupId>
+            <artifactId>jakarta.annotation-api</artifactId>
+            <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>jakarta.enterprise</groupId>
+            <artifactId>jakarta.enterprise.cdi-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.inject</groupId>
+            <artifactId>jakarta.inject-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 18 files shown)


fabiobrz (commented):
Thanks for this PR @pferraro. I am wondering whether this is enough, since we should actually bump the MicroProfile spec version in the POM to 7, see https://github.com/pferraro/eap-microprofile-test-suite/blob/master/pom.xml#L52.
This change was committed [recently](https://github.com/jboss-eap-qe/eap-microprofile-test-suite/commit/1f8af3c14187e775eb9a7f1f1d65c05ebe81218c#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L52), so a rebase on upstream master branch would do.

Then we need to see whether the tests actually pass, which can be done via a manual workflow execution..

pferraro:
> Thanks for this PR @pferraro. I am wondering whether this is enough, since we should actually bump the MicroProfile spec version in the POM to 7, see https://github.com/pferraro/eap-microprofile-test-suite/blob/master/pom.xml#L52. This change was committed [recently](https://github.com/jboss-eap-qe/eap-microprofile-test-suite/commit/1f8af3c14187e775eb9a7f1f1d65c05ebe81218c#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L52), so a rebase on upstream master branch would do.
> 
> Then we need to see whether the tests actually pass, which can be done via a manual workflow execution..

I explicitly omitted that version bump from the commit since it breaks other tests in the suite (as those modules are not yet MP7 ready).
Rather, what can do is override the version of the MP OpenAPI spec for this test module only.

pferraro:
> This change was committed [recently](https://github.com/jboss-eap-qe/eap-microprofile-test-suite/commit/1f8af3c14187e775eb9a7f1f1d65c05ebe81218c#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L52), so a rebase on upstream master branch would do.

Oops. I missed that vital detail the first time...

